PR #30, opened 2015-08-16
Fix bug when moving a window to the current workspace

pipcet:
* exwm-workspace.el (exwm-workspace-move-window): Run
          reparenting code when moving a window to the current workspace.

Hi! I think I've found a small bug when moving a window that's on another workspace to the current workspace with exwm-workspace-move-window: the test

    (when (/= exwm-workspace-current-index index)

was probably intended to test whether the window's current frame is changing, not whether we're currently on the workspace the window is moving to.


ch11ng:
Hi @pipcet,
`exwm-workspace-move-window` was designed to move window to other workspaces from the current one. If you want to make it working as you suggested, please fix the remaining code in that function (like buffer renaming, for example).

PR #32, opened 2015-08-16
Robustness

pipcet:
Hi,
I've been working on support for multiple top-level frames visible at the same time (since each of those frames has its own minibuffer, this effectively allows us to use the minibuffer space when running a client in one of those frames); I'll push my work to my master branch when I'm done with the PRs.

Anyway, it's happened to me a few times that functions signalled hard errors and broke things, so I think it's in the interest of code robustness to be a bit more lenient when we get unknown arguments.

For example, when killing clients, we would throw an error if the client had already died in the meantime; or, when using the interactive commands to set a client to full-screen mode, exwm-input--release-keyboard will be called with an invalid argument and the error breaks things.

So I think these changes would be a better idea than hunting down all callers that might supply incomplete arguments.

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -56,26 +56,27 @@ It's updated in several occasions, and only used by `exwm-input--set-focus'.")
 
 (defun exwm-input--set-focus (id)
   "Set input focus to window ID in a proper way."
-  (with-current-buffer (exwm--id->buffer id)
-    (if (and (not exwm--hints-input)
-             (memq xcb:Atom:WM_TAKE_FOCUS exwm--protocols))
-        (progn
-          (exwm--log "Focus on #x%x with WM_TAKE_FOCUS" id)
-          (xcb:+request exwm--connection
-              (make-instance 'xcb:icccm:SendEvent
-                             :destination id
-                             :event (xcb:marshal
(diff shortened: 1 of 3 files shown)


ch11ng:
Hi, @pipcet,
Thank you for your contributing.

> I've been working on support for multiple top-level frames visible at the same time (since each of those frames has its own minibuffer, this effectively allows us to use the minibuffer space when running a client in one of those frames); I'll push my work to my master branch when I'm done with the PRs.

What would it be like? Could you describe it in more detail?

I agree with you that most functions here shall be made more user-friendly. There is much more work needed I think. Please adapt your code according to the following guides before I can merge it:
* Properly use `if`, `when` and `unless` to make code more readable
* Combine adjacent tests to reduce indentation
* Use the more CLOS-like functions instead of those EIEIO specific (e.g. `slot-value` instead of `oref`) if possible

pipcet:
Of course, thanks for pointing those out! I tried fixing them, but if I missed something, please let me know. Would you prefer pull requests rebased into a single commit or are multiple commits okay?

ch11ng:
> Would you prefer pull requests rebased into a single commit or are multiple commits okay?

It's better to rebase multiple commits into a single one, as long as they are for a same purpose.

Issue #39, opened 2015-08-18
Use of `read-event`


pipcet:
I'm very suspicious of the use of `read-event` in 

    (defun exwm--make-emacs-idle-for (seconds)
      "Put Emacs in idle state for SECONDS seconds."
      (with-timeout (seconds) (read-event)))

It's causing problems when it is called from a buffer switch hook which is itself called in the middle of processing events. I also think we actually dropped those events that we did read.
    
In my experiments, I've been seeing errors like the following happen a lot with EXWM when switching back and forth between frames with the mouse pointer:
    
        Debugger entered--Lisp error: (wrong-type-argument listp #<frame *GNU Emacs* 0x4718c00>)
          handle-select-window((switch-frame #<frame *GNU Emacs* 0x4718c00>))
          funcall-interactively(handle-select-window (switch-frame #<frame *GNU Emacs* 0x4718c00>))
          call-interactively(handle-select-window nil nil)
          command-execute(handle-select-window)
    
What's happening, as far as I can work out, is that we're in the middle of reading an event from the keyboard when a buffer switch happens, and the buffer switch hook calls exwm code which calls `exwm--make-emacs-idle-for`, which in turn calls `read-event` and causes Emacs to overwrite some of the global event data.

Can we change that code to use either `sit-for` or `sleep-for`? Neither causes problems for me, and `sit-for` properly saves the events that it does read.

Of course it's best to avoid synchronous waits entirely, but I understand that might not always be possible.

ch11ng:
The intention of `exwm--make-emacs-idle-for` is to put the Emacs command loop into idle state and create opportunity for Emacs to carry out other tasks like finishing its UI initialization. `read-event` is the only way I can think of to make Emacs idle; `sit-for` and `sleep-for` merely freeze Emacs I'm afraid. Perhaps `exwm--make-emacs-idle-for` is abused and is incorrectly used in this case. Maybe we can fix it by putting the event read back to `unread-command-event`.

pipcet:
Hmm. I don't think it actually works that way, to be honest. `exwm--make-emacs-idle-for` will exit as soon as the first event is read, which I found a little confusing. `sit-for` shouldn't freeze emacs, though! Maybe that's something that only happens with an input method?

FWIW, here's the relevant code from `sit-for`:

    (let ((read (let ((input-method-function nil))
                  (read-event nil t seconds))))
      (or (null read)
	  (progn
	    (push (cons t read) unread-command-events)
	    nil))))))

[...]

PR #42, opened 2015-08-23
Avoid using the "no window manager" code in Emacs

pipcet:
This is as discussed on #39, I think. I'm using xcb:RECTANGLE as a geometry type rather than a vector; do let me know if you prefer using an [x y width height] vector and I'll change the code.

There's also some code to support horizontal/vertical resizing that needs some explanation: my setup is to have two "workspace" frames side-by-side, so I would set `keep-x`; however, in order to inhibit width changes when setting the font, I'd still like the fullscreen property to be fullboth.

        * exwm.el (exwm--on-ClientMessage): Handle fullscreen requests
	for frames.
	(exwm-init): Initialize workspaces after unlocking events.

	* exwm-workspace.el (exwm-workspace--init): Create frames as
	invisible, then make them visible only once their OverrideRedirect
	property has been set.

	* exwm-randr.el (exwm-randr--refresh): New frame parameter
	`exwm-geometry'.

	* exwm-layout.el (exwm-layout--fullscreen-geometry): New function.
	(exwm-layout--set-frame-fullscreen): New function.

The Emacs code is buggy, see https://github.com/ch11ng/exwm/issues/39

### DIFF
--- a/exwm-layout.el
+++ b/exwm-layout.el
@@ -162,6 +162,34 @@
     (setq exwm--fullscreen nil)
     (exwm-input-grab-keyboard)))
 
+;; This function is superficially similar to `exwm-layout-set-fullscreen', but
+;; they do very different things: `exwm-layout--set-frame-fullscreen' resizes a
+;; frame to the actual monitor size, `exwm-layout-set-fullscreen' resizes an X
+;; window to the frame size.
+(defun exwm-layout--set-frame-fullscreen (frame)
+  "Make frame FRAME fullscreen, with regard to its XRandR output if applicable."
+  (let ((geometry (or (frame-parameter frame 'exwm-geometry)
+                      (xcb:+request-unchecked+reply
+                          exwm--connection
+                          (make-instance 'xcb:GetGeometry
+                                         :drawable exwm--root))
+                      (make-instance 'xcb:RECTANGLE :x 0 :y 0
+                                     :width (x-display-width)
+                                     :height (x-display-height))))
+         (id (frame-parameter frame 'exwm-outer-id)))
+    (with-slots (x y width height) geometry
+                (xcb:+request exwm--connection
+                    (make-instance 'xcb:ConfigureWindow
+                                   :window id
+                                   :value-mask (logior xcb:ConfigWindow:X
+                                                       xcb:ConfigWindow:Y
+                                                       xcb:ConfigWindow:Width
+                                                       xcb:ConfigWindow:Height)
+                                   :x x :y y
+                                   :width width
+                                   :height height))
+                (xcb:flush exwm--connection))))
+
 (defun exwm-layout--refresh ()
   "Refresh layout."
   (let ((frame (selected-frame))
--- a/exwm-randr.el
+++ b/exwm-randr.el
@@ -85,6 +85,12 @@
           (setq geometry default-geometry
                 output nil))
         (set-frame-parameter frame 'exwm-randr-output output)
+        (set-frame-parameter frame 'exwm-geometry
+                             (make-instance 'xcb:RECTANGLE
+                                            :x (elt geometry 0)
+                                            :y (elt geometry 1)
+                                            :width (elt geometry 2)
+                                            :height (elt geometry 3)))
         (set-frame-parameter frame 'exwm-x (elt geometry 0))
         (set-frame-parameter frame 'exwm-y (elt geometry 1))
         (xcb:+request exwm--connection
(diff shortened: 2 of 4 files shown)


ch11ng:
Thank you for contributing!

> I'm using xcb:RECTANGLE as a geometry type rather than a vector; do let me know if you prefer using an [x y width height] vector and I'll change the code.

I almost forget about it! We should certainly prefer the more readable `xcb:RECTANGLE`.

Here are a few suggestions / questions about this PR:
* Move the condition in `cond` out into `when` since we only support `_NET_WM_STATE_ADD`
* No need to `push` to `props-new`; just feed the `data` slot a literal vector
* Do we still need to call `exwm--make-emacs-idle-for`?
* Is `exwm-layout--fullscreen-geometry` planed for other uses? If not, I think it's short enough to get embedded into `exwm-layout--set-frame-fullscreen`
* What's the rationale behind the `xcb:SendEvent` in `exwm-layout--set-frame-fullscreen`? It seems working without this request

pipcet:
Thank you, I've fixed all of those, I think. `(frame-parameter (selected-frame) 'fullscreen)` is correct now.

Can you tell me more about the other call to `exwm--make-emacs-idle-for`, in `exwm-workspace-switch`? Why is it necessary?

ch11ng:
I believe the `exwm--make-emacs-idle-for` in `exwm-workspace-switch` has something to do with RandR. If I recall correctly, without this form the input focus won't change when user click on a inactive workspace frame (on another monitor). But I have no extra monitor to validate it right now.

pipcet:
> There's also some code to support horizontal/vertical resizing that needs some explanation: my setup is to have two "workspace" frames side-by-side, so I would set `keep-x`; however, in order to inhibit width changes when setting the font, I'd still like the fullscreen property to be fullboth.

Nevermind about that, actually, it's needless complication. I'll remove `keep-x` and `keep-y`, if you don't mind.

> If I recall correctly, without this form the input focus won't change when user click on a inactive workspace frame (on another monitor). But I have no extra monitor to validate it right now.

Oh, I can reproduce a small problem without the `exwm--make-emacs-idle-for` call: the pointer is warped by `select-frame-set-input-focus` before the mouse release event happens, so some text is selected but shouldn't be. Of course, the `exwm--make-emacs-idle-for` call only works when the click and the release happen within .1 s of each other, so that's not a good solution anyway. Is it possible that was the issue that required the call?

ch11ng:
> Is it possible that was the issue that required the call?

Well, very likely, but again I cannot confirm it right now.

ch11ng:
This PR introduces an issue that on startup EXWM will pick a wrong workspace as the active one.

And also:
* There are some extra whitespaces introduced in [this line](https://github.com/pipcet/exwm/blob/ewmh-fullscreen/exwm-layout.el#L171)
* I meant move [this condition](https://github.com/pipcet/exwm/blob/ewmh-fullscreen/exwm.el#L449) outside `dolist`; there is no need to traverse the list if this it's not satisfied

pipcet:
> This PR introduces an issue that on startup EXWM will pick a wrong workspace as the active one.

Thanks for spotting that! Is that permanent for you, or do you mean that each frame is briefly selected as it is mapped and raised by GTK? It's hard to reproduce here.

ch11ng:
> > This PR introduces an issue that on startup EXWM will pick a wrong workspace as the active one.

> Thanks for spotting that! Is that permanent for you, or do you mean that each frame is briefly selected as it is mapped and raised by GTK? It's hard to reproduce here.

The workspace is switched multiple times on startup (in `focus-in-hook` I believe), as is clearly shown in log.

pipcet:
I believe it's `buffer-list-update-hook`. It'd be great to know whether it still happens with the latest version—unfortunately, I suspect this bug depends on how fast the X server processes requests...

ch11ng:
I can't really say where the problem is, but the workspace is switched in `focus-in-hook`:

```
[EXWM] Refresh workspace #<frame *scratch* 0x11c06e0>
[EXWM] Refresh workspace #<frame *scratch* 0x355b308>
[EXWM] Refresh workspace #<frame *scratch* 0x36ac410>
[EXWM] Refresh workspace #<frame *scratch* 0x377e700>
[EXWM] Focus on workspace 0
[EXWM] Focus on workspace 2
[EXWM] Workspace was switched unexpectedly
[EXWM] Focus on workspace 3
[EXWM] Workspace was switched unexpectedly
[EXWM] Focus on workspace 0
[EXWM] Workspace was switched unexpectedly
[EXWM] Focus on workspace 0
[EXWM] Focus on #<window 3 on *scratch*>
[EXWM] Focus on #<window 9 on *scratch*>
[EXWM] Focus on workspace 3
[EXWM] Workspace was switched unexpectedly
[EXWM] Update modifier mapping
```

pipcet:
Ah, thanks for the log. You're right, here's what happened: when GTK shows a window, it automatically raises it, so unless we're fast enough to lower it again, an EnterNotify is generated, which generated a FOCUS_IN_EVENT because the focus hadn't been previously set.

Fixed now, I hope.

ch11ng:
Applied. Thanks!

PR #45, opened 2015-08-25
Warp mouse pointer only when necessary.

pipcet:
See https://github.com/ch11ng/exwm/issues/39

Feel free to leave this open until you can test on XRandR, of course, there might have been another problem I haven't thought of.

This version checks the mouse pointer's absolute position is over the frame, not whether its

### DIFF
--- a/exwm-floating.el
+++ b/exwm-floating.el
@@ -258,7 +258,7 @@
       (with-slots (root-x root-y win-x win-y)
           (xcb:+request-unchecked+reply exwm--connection
               (make-instance 'xcb:QueryPointer :window id))
-        (select-frame-set-input-focus frame) ;raise and focus it
+        (exwm-workspace--select-frame frame)
         (setq width (frame-pixel-width frame)
               height (frame-pixel-height frame))
         (unless type
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -117,13 +117,12 @@ It's updated in several occasions, and only used by `exwm-input--set-focus'.")
           (progn
             (when exwm--floating-frame
               (redirect-frame-focus exwm--floating-frame nil)
-              (select-frame-set-input-focus exwm--floating-frame t))
+              (exwm-workspace--select-frame exwm--floating-frame t))
             (exwm--log "Set focus on #x%x" exwm--id)
             (exwm-input--set-focus exwm--id))
         (when (eq (selected-window) exwm-input--focus-window)
           (exwm--log "Focus on %s" exwm-input--focus-window)
-          (select-frame-set-input-focus (window-frame exwm-input--focus-window)
-                                        t)
+          (exwm-workspace--select-frame (window-frame exwm-input--focus-window) t)
           (dolist (pair exwm--id-buffer-alist)
             (with-current-buffer (cdr pair)
               (when (and exwm--floating-frame
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -88,10 +88,32 @@
 (defvar exwm-workspace--current nil "Current active workspace.")
 (defvar exwm-workspace-current-index 0 "Index of current active workspace.")
 
-(defun exwm-workspace-switch (index &optional force)
+(defun exwm-workspace--select-frame (frame &optional norecord)
+  ;; Move mouse when necessary
+  (let* ((position (mouse-pixel-position))
+         (x (+ (cadr position) (car (frame-position (car position)))))
+         (y (+ (cddr position) (cdr (frame-position (car position)))))
+         (edges (frame-edges frame)))
+    (if (or (not (memq frame exwm-workspace--list))
+            (and x
+                 (>= x (nth 0 edges))
+                 (> (nth 2 edges) x)
+                 (>= y (nth 1 edges))
+                 (> (nth 3 edges) y)))
+        (progn (select-frame frame)
+               (raise-frame frame)
+               (x-focus-frame frame))
+      (select-frame-set-input-focus frame norecord)
+      (when (or (>= x (nth 2 edges)) (>= y (nth 3 edges)))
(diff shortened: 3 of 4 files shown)


ch11ng:
This PR causes a horrible issue that when you switch workspace quite fast (keep typing `s-1`, `s-2` and `s-3` for some times for instance), you will end up with a situation where the input focus is not stable.

Also consider properly prefix `exwm--select-frame` (e.g. with `exwm-workspace` if it's put in exwm-workspace.el).

pipcet:
Can reproduce that here. It's another old bug that was hidden by the way `exwm--make-emacs-idle-for` discarded evens: It's spectacularly useless of Emacs, but `focus-in-hook` is called *before* (selected-frame) is updated, so we switched back to the old frame...

I see only two ways around that, and neither is particularly pretty: a timer, or advising handle-switch-frame. Maybe you have a better idea?

ch11ng:
> focus-in-hook is called before (selected-frame) is updated

It sounds like `focus-in-hook` is basically useless.

> I see only two ways around that, and neither is particularly pretty: a timer, or advising handle-switch-frame. Maybe you have a better idea?

We still have X events.

pipcet:
Indeed. X events is what the somewhat-outdated code on my master branch uses, but then it's meant not to have a concept of the "current" workspace at all.

But do we really need to call x-focus-frame at all after a focus-in event? I think the focus-in event is telling us that it has already happened; we might still have to raise the frame (though Emacs does that if the frame is auto-raise) and warp the mouse pointer, but we don't need to select-frame (it's already selected) or x-focus-frame (it should already have focus).

ch11ng:
I have no idea if `select-frame` and `x-focus-frame` are really necessary here, but sometimes it seems we cannot set input focus to a frame without them. Input focus is what I feel the most difficult part in building EXWM. Maybe the current implementation can be simplified.

pipcet:
To be honest, I don't even know whether EXWM is a click-to-focus window manager or a focus-follows-mouse (sloppily) window manager. Ideally, it would be both, depending on the values of `focus-follows-mouse` and `mouse-autoselect-window`.

I'm a huge focus-follows-mouse fan, but hate cursor warping. I'm quite upset that Emacs uses the same variable for "focus follows mouse" and "mouse follows focus", because I want one and not the other.

But back to EXWM:
I'd really like to investigate those situations. I suspect they were caused by the old `exwm--make-emacs-idle-for` code. Remember that that code (which is still in your master branch) eats *random* events, which we then lose. That might be a focus-in event, and then the focus in would be ignored.

So I think once I fix up these two PRs, all our problems should just go away.

ch11ng:
> To be honest, I don't even know whether EXWM is a click-to-focus window manager or a focus-follows-mouse (sloppily) window manager.

Click-to-focus, since its birth. Emacs itself is also click-to-focus I think (considering Emacs windows on a same frame). But it's not so important since input focus is usually changed with keyboard.

> So I think once I fix up these two PRs, all our problems should just go away.

Good luck! But do be careful with floating windows; most input focus issues I experienced are related to them.

pipcet:
Emacs can be customized to be either click-to-focus or focus-follows-mouse by those two variables I mentioned.

Thanks for your advice, it's really good to know what to focus on.

uu1101:
I feel the same w.r.t. focus-follows-mouse: EXWM behaves in a hybrid way with my settings (`focus-follows-mouse` is `nil`, `mouse-autoselect-window` is `t`).  I guess it no longer makes sense for these two to have different values, as Emacs is now the window manager.

With regards to the pull request, I'm trying it out, as I have some changes to the workspace management where I can see a similar issue (unpredictably input focus goes nuts, no longer possible to switch worspaces).  Building latest emacs, as `mouse-absolute-pixel-position` has only been introduces a few days ago.

Thank you both for your efforts!

uu1101:
As far as I can tell this works like a charm.  Thank you!

pipcet:
@uu1101 Thanks for the warning about the function having been introduced only recently. That needs fixing, of course!

pipcet:
@uu1101 Thanks for the warning about the function having been introduced only recently. That needs fixing, of course! And thanks a lot for testing, it's great that it's working in practice, at least (though there's still at least one known bug, so we can't quite merge this yet).

pipcet:
I've tested this version some more, and I think it's good to go; however, I think `exwm-input--on-focus-in` also assumes it is being called with the new frame as `(selected-frame)`, which doesn't work, so maybe we should fix that, too.

pipcet:
We really, really, need to get rid of exwm--make-emacs-idle-for. I've just come across this backtrace:
https://gist.github.com/pipcet/d745fa6b44274d9dbce6

What's happening is that our process handler gets called by exwm--make-emacs-idle-for, but the timeout hits, and the event never gets handled and we end up with the process in stuck state, but only if the X server takes more than 0.1 seconds to process a request.

ch11ng:
Yes, this is problematic, and we have agreed to remove `exwm--make-emacs-idle-for`.

pipcet:
Thank you, that wasn't totally clear to me.

Is this PR taking the right approach at all or do you want me to try doing it another way?

ch11ng:
I cannot tell whether the approach you take in this PR will work eventually or not. Indeed, I don't think I have a thorough understanding of the focus switch mechanism of Emacs frames. I don't care about in which way the PR is implemented (of course I'd prefer the neat ways) as long as it solves some problem.

ch11ng:
Since 9c95c03e18f6d5cf78bcd54bf00f8055a3863336 the pointer is no longer warped.  The problem was: when a global key grab was activated, the input focus would transfer to the root window, and then to the workspace containing the pointer.  By grabbing key on the new workspace container, there's no such problem anymore.

Issue #40, opened 2015-08-19
minibuffer-setup-hook vs window-size-change-functions


pipcet:
Running

    (progn
      (message (make-string 3000 ?*))
      (start-process "xterm" nil "xterm"))

produces a small problem: the minibuffer gets resized without our noticing. The problem is that `minibuffer-setup-hook` doesn't really say anything about the minibuffer being resized, just about its being entered.

I think this is also the problem in the screenshot at #27.

There are two alternatives: `window-size-change-functions` is precisely what we want, but it's only called when there is a menu bar. `window-text-change-functions` is called whenever anything might have changed, which is very often, so it might lead to noticeable slowdown.

There's also `echo-area-clear-hook`, but that's not called when the echo area grows, just when it shrinks.

I think the best way forward is to fix the xdisp.c code to call `window-size-change-functions` even when there is no menu bar, and submit that to upstream Emacs. Until it's fixed, things aren't all that horrible, but slightly annoying.

Or do you have a better idea?

pipcet:
Sorry, I must have gotten confused between `prepare_menu_bars` and `update_menu_bar`. `window-size-change-functions` is what we want.

ch11ng:
> Running
>
> ```
(progn
  (message (make-string 3000 ?*))
  (start-process "xterm" nil "xterm"))
```
> produces a small problem: the minibuffer gets resized without our noticing. The problem is that `minibuffer-setup-hook` doesn't really say anything about the minibuffer being resized, just about its being entered.

Rather than *minibuffer*, I think it's *echo area*. They are different things. `minibuffer-setup-hook` certainly has nothing to do with it.

> There are two alternatives: `window-size-change-functions` is precisely what we want, but it's only called when there is a menu bar.
> ...
> I think the best way forward is to fix the xdisp.c code to call `window-size-change-functions` even when there is no menu bar, and submit that to upstream Emacs.

I haven't checked the C source yet, but what's the reason for such inconsistent behaviors? If it's a result of carelessness, it should be fixed of course.

> Or do you have a better idea?

We originally only do refresh in `window-configuration-change-hook`; `minibuffer-setup-hook` was added later to account for minibuffer resize. Now we want to add a third. It's better to do refresh just in one place if possible.

pipcet:
Again, I was wrong about `window-size-change-functions`; it's run when there is no menu bar, but there still appear to be cases when it isn't run even though the window sizes do change.

You're right, of course, about the difference between minibuffer and echo area.

[...]

PR #47, opened 2015-08-26
Use pre-redisplay-function to detect mini-window resizes.

pipcet:
Fixes https://github.com/ch11ng/exwm/issues/40

See http://debbugs.gnu.org/cgi/bugreport.cgi?bug=21333

Do let me know what needs changing. In particular, do you prefer a simple wrapper around `exwm-layout--refresh` to making it accept and ignore arguments?

I'd also like to thank you for your cooperation in general and the xterm fix in particular. I confess I never previously understood how XAllowEvents worked!

(I'd originally rewritten much more of the code in exwm-layout.el, but I never really got that to work the way it should.)

### DIFF
--- a/exwm-layout.el
+++ b/exwm-layout.el
@@ -189,8 +189,8 @@
                          :height height))
       (xcb:flush exwm--connection))))
 
-(defun exwm-layout--refresh ()
-  "Refresh layout."
+(defun exwm-layout--refresh (&rest args)
+  "Refresh layout.  ARGS are ignored to run in abnormal hooks."
   (let ((frame (selected-frame))
         (placeholder (get-buffer "*scratch*"))
         windows)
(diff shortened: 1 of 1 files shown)


ch11ng:
I think `pre-redisplay-function` is called too often; it's even called when I just sit there doing nothing.

On the other hand, I have no idea of how to properly use `pre-redisplay-function`, but the use of advice is discouraged according to the Elisp coding convention (I do use advice in one place but that's not enabled by default).

pipcet:
You're right, of course! I removed that bit from my rewrite and forgot to add it back in. It's called very often and we absolutely must filter out those calls that aren't for us. Sorry about that.

There's no way around the advice, unfortunately, since it's a single function, not a list of functions or a proper hook. That's also something the Emacs people are discussing, though.

ch11ng:
The issue was resolved by 0e4055d3392537cd4138181b9c615fa03053add8, so I'm closing this PR.

PR #49, opened 2015-08-27
Dynamic workspaces

uu1101:
These changes remove the 10 workspaces limitation, and allow for a more natural workspace workflow: creating a frame creates a workspace, deleting a frame deletes the workspace.

The workspace switcher has been reworked to allow switching to any multi-digit workspaces.  (Nevertheless, at some point it should probably be extracted into a exwm-pager.el that does not rely so much on the internal, and proofs the interaction with other pagers.)

Users are able to start with a pre-defined number of workspaces by creating as many frames as they like before starting EXWM.

Right now, new workspaces are appended at the end, and windows in a deleted workspace are moved to the next one when it is destroyed.  This could be made configurable, but I think it can wait.

Please, note that #45 should be applied first, as the issues it fixes wreak havoc when manipulating workspaces.

The configuration example in the Wiki should probably be updated as well.

I have left more-or-less small commits, in case it helps review. Of course, I appreciate any comment, critique, etc.

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -28,6 +28,11 @@
 
 (defvar exwm-workspace-number 4 "Number of workspaces (1 ~ 10).")
 (defvar exwm-workspace--list nil "List of all workspaces (Emacs frames).")
+
+(defun exwm-workspace--count ()
+  "Retrieve total number of workspaces."
+  (length exwm-workspace--list))
+
 (defvar exwm-workspace--switch-map
   (let ((map (make-sparse-keymap)))
     (define-key map [t] (lambda () (interactive)))
@@ -41,7 +46,7 @@
     (define-key map "\C-a" (lambda () (interactive) (goto-history-element 1)))
     (define-key map "\C-e" (lambda ()
                              (interactive)
-                             (goto-history-element (length exwm-workspace--list))))
+                             (goto-history-element (exwm-workspace--count))))
     (define-key map "\C-g" 'abort-recursive-edit)
     (define-key map "\C-]" 'abort-recursive-edit)
     (define-key map "\C-j" 'exit-minibuffer)
@@ -61,7 +66,7 @@
 
 (defun exwm-workspace--update-switch-history ()
   "Update the history for switching workspace to reflect the latest status."
-  (let* ((num (length exwm-workspace--list))
+  (let* ((num (exwm-workspace--count))
          (sequence (number-sequence 0 (1- num)))
          (not-empty (make-vector num nil)))
     (dolist (i exwm--id-buffer-alist)
@@ -105,7 +110,7 @@ The optional FORCE option is for internal use only."
                      . ,(1+ exwm-workspace-current-index)))))
         (cl-position idx exwm-workspace--switch-history :test 'equal)))))
   (when index
-    (unless (and (<= 0 index) (< index (length exwm-workspace--list)))
+    (unless (and (<= 0 index) (< index (exwm-workspace--count)))
       (user-error "[EXWM] Workspace index out of range: %d" index))
     (when (or force (/= exwm-workspace-current-index index))
       (let ((frame (elt exwm-workspace--list index)))
@@ -164,7 +169,7 @@ The optional FORCE option is for internal use only."
                    . ,(1+ exwm-workspace-current-index)))))
       (cl-position idx exwm-workspace--switch-history :test 'equal))))
   (unless id (setq id (exwm--buffer->id (window-buffer))))
-  (unless (and (<= 0 index) (< index (length exwm-workspace--list)))
+  (unless (and (<= 0 index) (< index (exwm-workspace--count)))
     (user-error "[EXWM] Workspace index out of range: %d" index))
   (with-current-buffer (exwm--id->buffer id)
     (let ((frame (elt exwm-workspace--list index)))
@@ -215,7 +220,7 @@ The optional FORCE option is for internal use only."
 (defun exwm-workspace--add-frame-as-workspace (frame)
   "Configure frame FRAME to be treated as a workspace."
   (cond
-   ((>= (length exwm-workspace--list) exwm-workspace-number)
+   ((>= (exwm-workspace--count) exwm-workspace-number)
     (delete-frame frame)
     (user-error "[EXWM] Too many workspaces: maximum is %d" exwm-workspace-number))
    ((memq frame exwm-workspace--list)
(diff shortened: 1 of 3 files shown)


exwm-workspace.el
```diff
@@ -211,38 +252,26 @@ The optional FORCE option is for internal use only."
...
-          (outer-id (string-to-int (frame-parameter i 'outer-window-id))))
+(defun exwm-workspace--add-frame-as-workspace (frame)
+  "Configure frame FRAME to be treated as a workspace."
+  (cond
+   ((memq frame exwm-workspace--list)
+    (exwm--log "Frame is already a workspace: %s" frame))
+   ((not (display-graphic-p frame))
+    (exwm--log "Frame is not graphical: %s" frame))
+   ((not (display-graphic-p frame))
+    (exwm--log "Frame is not graphical: %s" frame))
```
pipcet: You appear to be checking (display-graphic-p frame) twice here.

exwm-workspace.el
```diff
@@ -252,26 +281,106 @@ The optional FORCE option is for internal use only."
...
-  ;; Set _NET_VIRTUAL_ROOTS
-  (xcb:+request exwm--connection
-      (make-instance 'xcb:ewmh:set-_NET_VIRTUAL_ROOTS
-                     :window exwm--root
-                     :data (vconcat (mapcar
-                                     (lambda (i)
-                                       (frame-parameter i 'exwm-window-id))
-                                     exwm-workspace--list))))
+  ;; Make new frames create new workspaces.
+  (setq window-system-default-frame-alist '((x . ((visibility . nil)))))
```
pipcet: I think it might be better not to overwrite the existing alist if there is one, just to change the visibility property.

exwm-workspace.el
```diff
@@ -252,26 +281,106 @@ The optional FORCE option is for internal use only."
...
-    (set-frame-parameter i 'visibility t)
-    (lower-frame i)
-    (set-frame-parameter i 'fullscreen 'fullboth))
-  (raise-frame (car exwm-workspace--list))
+                         :event-mask xcb:EventMask:SubstructureRedirect)))
+    (xcb:flush exwm--connection)
+    ;; We have to delay making the frame visible until the
+    ;; override-redirect flag has been set.
+    (set-frame-parameter frame 'visibility t)
+    (lower-frame frame)
```
pipcet: Do we really want this `lower-frame`? We need to lower the initial workspaces so the first one ends up on top, but for newly-created frames I think it's more intuitive for the new frame to be topmost.

exwm-workspace.el
uu1101: You are right.  I would appreciate it if you have some suggestion on how to do that on a non convoluted way, as my elisp is somewhat...

exwm-workspace.el
uu1101: Oops! I have pushed a fix.

exwm-workspace.el
uu1101: To be honest... I don't know.  I have tested commenting the line and see no difference neither w.r.t. the initial workspaces nor when adding/removing... Shall I drop it?

uu1101:
@pipcet Thank you very much for your review!

exwm-workspace.el
pipcet: Not my decision to make, but I think you should, and call `exwm--select-frame` instead. That would make our idea of which frame we're on consistent, I think.

pipcet:
@uu1101 This code is great, sorry for being so nit-picky.

One last thing:
Is it possible the code for removing windows from a deleted workspace isn't in the PR? I don't see any code that does that, and it doesn't appear to work here (Iceweasel segfaults when its frame is deleted).

uu1101:
Ah! No, that should be the default X behaviour, see section on `_NET_NUMBER_OF_DESKTOPS`: 

>  If the number of desktops is shrinking and _NET_CURRENT_DESKTOP is out of the new range of available desktops, then this MUST be set to the last available desktop from the new set. Clients that are still present on desktops that are out of the new range MUST be moved to the very last desktop from the new set. For these _NET_WM_DESKTOP MUST be updated. 

http://standards.freedesktop.org/wm-spec/latest/ar01s03.html#idm139870830094832

What is done in that commit is making sure we are in some not-going-to-be-deleted workspace, so that all windows from the previous are moved there.

Does that make sense?

pipcet:
I think you're misreading that paragraph: it says *the window manager* MUST move all windows to the new virtual desktop before destroying the old one.

exwm-workspace.el
uu1101: I came up with the following:
~~~{.elisp}
  (if (assq 'x window-system-default-frame-alist)
      (push '(visibility . nil) (rassq 'x window-system-default-frame-alist))
    (setq window-system-default-frame-alist '((x . ((visibility . nil))))))
~~~
Is that idiomatic?

pipcet: I think you have an `rassq` for an `assq`, and the `setq` should be a push, but I might be wrong.

``` elisp
  (if (assq 'x window-system-default-frame-alist)
      (push '(visibility . nil) (cdr (assq 'x window-system-default-frame-alist)))
    (push '(x . ((visibility))) window-system-default-frame-alist))
```

maybe.

uu1101:
@pipcet:

Hmm... I see, you are right.  This needs to be fixed as well.

@ch11ng:

> We'd better keep this variable public so that users can use it to display the current workspace number in mode-line, for example.

Oh, sure, I wondered why was it public.  Reverting that.

> I've not tried this yet but does exwm-workspace--add-frame-as-workspace correctly handle floating frames? Also there is a typo.

I have to check floating frames, as I don't use them much.

If you refer to `=` with the typo, I think it's the "numeric equal" operator.

Thank you very much for your review!

---

This obviously needs much more work, primarily wanted to get it out there.  I'll be away for a couple of weeks and won't be able to work on it.  You might want to close the pr meanwhile.

ch11ng:
> If you refer to `=` with the typo, I think it's the "numeric equal" operator.

No, I meant `llength`.

uu1101:
Hello,

Thank you both for your review and comments.  @ch11ng, I've seen you plan to put this on ELPA.  I am glad you take that decision, especially this early.  Sadly, this means I have to abandon this contribution, as this handle is to remain pseudonym.  For what is worth, I'm releasing the above code in the public domain; I hope someone else can build on these ideas.

Nevertheless, I'll probably be able to contribute issues and small snippets.

Best regards,
uu1101.

ch11ng:
Hi @uu1101,
I think it will also be helpful if you can provide detailed descriptions of the features you want to implement, especially those not present in your code (if any). Anyway, thank you for your contribution.

ch11ng:
I've merged #172 instead which is a neat implementation.  Thanks anyway.

PR #51, opened 2015-08-28
Handle failed GrabPointer requests.

pipcet:
I've just seen this backtrace: https://gist.github.com/pipcet/46eb9b87d75660b7c330

So I think we need this patch, or maybe to change the xcb bindings.

Note that I'm using VNC, so it might just be VNC messing with the pointer somehow.

### DIFF
--- a/exwm-floating.el
+++ b/exwm-floating.el
@@ -234,6 +234,19 @@
 (defvar exwm-floating--moveresize-calculate nil
   "Calculate move/resize parameters [frame-id event-mask x y width height].")
 
+;; Test if the pointer can be grabbed
+(defun exwm-floating--grab-pointer (frame-id)
+  (let ((result (xcb:+request-unchecked+reply exwm--connection
+                    (make-instance 'xcb:GrabPointer
+                                   :owner-events 0 :grab-window frame-id
+                                   :event-mask xcb:EventMask:NoEvent
+                                   :pointer-mode xcb:GrabMode:Async
+                                   :keyboard-mode xcb:GrabMode:Async
+                                   :confine-to xcb:Window:None
+                                   :cursor xcb:Cursor:None
+                                   :time xcb:Time:CurrentTime))))
+    (and result (= (slot-value result 'status) xcb:GrabStatus:Success))))
+
 (defun exwm-floating--start-moveresize (id &optional type)
   "Start move/resize."
   (let ((buffer (exwm--id->buffer id))
@@ -241,19 +254,7 @@
     (when (and buffer
                (setq frame (with-current-buffer buffer exwm--floating-frame))
                (setq frame-id (frame-parameter frame 'exwm-outer-id))
-               ;; Test if the pointer can be grabbed
-               (= xcb:GrabStatus:Success
-                  (slot-value
-                   (xcb:+request-unchecked+reply exwm--connection
-                       (make-instance 'xcb:GrabPointer
-                                      :owner-events 0 :grab-window frame-id
-                                      :event-mask xcb:EventMask:NoEvent
-                                      :pointer-mode xcb:GrabMode:Async
-                                      :keyboard-mode xcb:GrabMode:Async
-                                      :confine-to xcb:Window:None
-                                      :cursor xcb:Cursor:None
-                                      :time xcb:Time:CurrentTime))
-                   'status)))
+               (exwm-floating--grab-pointer frame-id))
       (setq exwm--floating-edges nil)   ;invalid by now
       (with-slots (root-x root-y win-x win-y)
           (xcb:+request-unchecked+reply exwm--connection


ch11ng:
I thought such issue would never appear again. I think it's better to retrieve and log the error (and also the status if any) so that we may have a chance to fix it. Do you have any reliable way to reproduce this bug?

pipcet:
I can reproduce it, but not reliably, I'm afraid. I'll try getting a log with xcb debugging enabled, if that's what you mean by "log the error"?

ch11ng:
I suggest using `xcb:+request+reply`. Because there is no reply, there must be at least an error.

pipcet:
`xcb+request+reply` appears to have returned `(nil nil)`. Should that even be possible?

ch11ng:
I think it should never be possible. It sounds to be a XELB bug then. Consider turn on debug for XELB?

pipcet:
Messages:
```
[XELB LOG] Event received: [4 3 244 0 227 55 236 69 151 0 0 0 13 0 96 2 24 0 96 2 149 2 100 1 63 0 35 0 0 0 1 0]
[XELB LOG] Cache request: [eieio-class-tag--xcb:GrabPointer t t 26 0 6293849 0 1 1 0 0 0]
Entering debugger...
[EXWM] Refresh workspace #<frame *scratch* 0x318d1f0>
```

Backtrace:
```
Debugger entered--Lisp error: (wrong-type-argument (or eieio-object class) nil obj)
  signal(wrong-type-argument ((or eieio-object class) nil obj))
  slot-value(nil status)
  (= xcb:GrabStatus:Success (slot-value (car r) (quote status)))
  (let ((r (xcb:-+reply exwm--connection (xcb:-+request exwm--connection (make-instance (quote xcb:GrabPointer) :owner-events 0 :grab-window frame-id :event-mask xcb:EventMask:NoEvent :pointer-mode xcb:GrabMode:Async :keyboard-mode xcb:GrabMode:Async :confine-to xcb:Window:None :cursor xcb:Cursor:None :time xcb:Time:CurrentTime)) nil))) (= xcb:GrabStatus:Success (slot-value (car r) (quote status))))
  (and buffer (setq frame (save-current-buffer (set-buffer buffer) exwm--floating-frame)) (setq frame-id (frame-parameter frame (quote exwm-outer-id))) (let ((r (xcb:-+reply exwm--connection (xcb:-+request exwm--connection (make-instance (quote xcb:GrabPointer) :owner-events 0 :grab-window frame-id :event-mask xcb:EventMask:NoEvent :pointer-mode xcb:GrabMode:Async :keyboard-mode xcb:GrabMode:Async :confine-to xcb:Window:None :cursor xcb:Cursor:None :time xcb:Time:CurrentTime)) nil))) (= xcb:GrabStatus:Success (slot-value (car r) (quote status)))))
  (if (and buffer (setq frame (save-current-buffer (set-buffer buffer) exwm--floating-frame)) (setq frame-id (frame-parameter frame (quote exwm-outer-id))) (let ((r (xcb:-+reply exwm--connection (xcb:-+request exwm--connection (make-instance ... :owner-events 0 :grab-window frame-id :event-mask xcb:EventMask:NoEvent :pointer-mode xcb:GrabMode:Async :keyboard-mode xcb:GrabMode:Async :confine-to xcb:Window:None :cursor xcb:Cursor:None :time xcb:Time:CurrentTime)) nil))) (= xcb:GrabStatus:Success (slot-value (car r) (quote status))))) (progn (setq exwm--floating-edges nil) (let* ((object (car (xcb:-+reply exwm--connection (xcb:-+request-unchecked exwm--connection ...) nil)))) (progn (select-frame-set-input-focus frame) (setq width (frame-pixel-width frame) height (frame-pixel-height frame)) (if type nil (setq x (/ (* 3 ...) (float width)) y (/ (* 3 ...) (float height)) type (cond (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_TOPLEFT) (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_TOPRIGHT) (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_BOTTOMRIGHT) (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_BOTTOMLEFT) (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_TOP) (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_RIGHT) (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_BOTTOM) (... xcb:ewmh:_NET_WM_MOVERESIZE_SIZE_LEFT)))) (if type (progn (cond (... ...) (... ...) (... ...) (... ...) (... ...) (... ...) (... ...) (... ...) (... ...)) (car (xcb:-+reply exwm--connection ... nil))))))))
  (let ((buffer (exwm--id->buffer id)) frame frame-id cursor) (if (and buffer (setq frame (save-current-buffer (set-buffer buffer) exwm--floating-frame)) (setq frame-id (frame-parameter frame (quote exwm-outer-id))) (let ((r (xcb:-+reply exwm--connection (xcb:-+request exwm--connection ...) nil))) (= xcb:GrabStatus:Success (slot-value (car r) (quote status))))) (progn (setq exwm--floating-edges nil) (let* ((object (car (xcb:-+reply exwm--connection ... nil)))) (progn (select-frame-set-input-focus frame) (setq width (frame-pixel-width frame) height (frame-pixel-height frame)) (if type nil (setq x (/ ... ...) y (/ ... ...) type (cond ... ... ... ... ... ... ... ...))) (if type (progn (cond ... ... ... ... ... ... ... ... ...) (car ...))))))))
  exwm-floating--start-moveresize(39845901)
```

Value of `r` is `(nil nil)`, as above.

pipcet:
Sorry, that omitted the important bit. Here's a more complete log:

```
[XELB LOG] Event received: [4 3 244 0 227 55 236 69 151 0 0 0 13 0 96 2 24 0 96 2 149 2 100 1 63 0 35 0 0 0 1 0]
[XELB LOG] Cache request: [eieio-class-tag--xcb:GrabPointer t t 26 0 6293849 0 1 1 0 0 0]
Entering debugger...
[EXWM] Refresh workspace #<frame *scratch* 0x318d1f0>
[EXWM] Focus on #<window 9 on *Backtrace*>
[EXWM] Focus on workspace 3
[EXWM] Workspace was switched unexpectedly
[XELB LOG] Reply received: [1 0 245 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
```

ch11ng:
I think there's a bug in `xcb:-+reply` but I cannot tell where it actually is. It seems this function accidentally exited before the reply (the last line in your log I think, where `[245 0]` AKA 245 is the sequence number of the reply) was received. Most likely there's something wrong with [these lines](https://github.com/ch11ng/xelb/blob/c99266a3318c077ef4ef9ca6d26582be0af90b16/xcb.el#L514-L516). Could you please verify these sequence numbers: `sequence`, `reply-sequence` and `error-sequence`?

pipcet:
I think what's happening is that we have recursive requests:
```
[XELB LOG] Cache request: [eieio-class-tag--xcb:GrabButton t t 28 0 6294024 4 0 1 0 0 0 1 32768]
[XELB LOG] Cache request: [eieio-class-tag--xcb:ewmh:get-_NET_WM_NAME t t 20 0 6294024 xcb:Atom:_NET_WM_NAME xcb:Atom:UTF8_STRING 0 1000000000]
[EXWM] Focus on #<window 3 on c-finalizer.el>
[XELB LOG] Reply received: [1 24 21 1 0 0 0 0 151 0 0 0 116 2 41 1 168 1 199 1 0 0 0 0 0 0 0 0 0 0 0 0]
[XELB LOG] Reply received: [1 8 24 1 3 0 0 0 72 1 0 0 0 0 0 0 11 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 80 105 99 107 32 97 32 102 111 110 116 0]
[XELB LOG] Cache request: [eieio-class-tag--xcb:icccm:get-WM_NAME t t 20 0 6294024 xcb:Atom:WM_NAME xcb:GetPropertyType:Any 0 1000000000]
[XELB LOG] Cache request: [eieio-class-tag--xcb:icccm:get-WM_TRANSIENT_FOR t t 20 0 6294024 xcb:Atom:WM_TRANSIENT_FOR xcb:Atom:WINDOW 0 1]
[XELB LOG] Reply received: [1 8 25 1 3 0 0 0 31 0 0 0 0 0 0 0 11 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 80 105 99 107 32 97 32 102 111 110 116 0]
[XELB LOG] Reply received: [1 32 26 1 1 0 0 0 33 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 20 0 96 0]
```

My theory is this:

The problem is that we allow handlers to be called recursively:
 - something calls xcb:-request
 - xcb:-request calls accept-process-output
 - accept-process-output calls process filter
 - process filter calls an event handler
 - the other handler calls xcb:-request AGAIN
 - the response to the inner request arrives before the response to the outer request does
 - the event handler exits
 - control returns to xcb:-request
 - the sequence number has increased, but the response to the outer request hasn't arrived yet.
 - our bug occurs.

The "inner" request gets a higher sequence number than the "outer" request, the response to the inner request arrives before the response to the outer request, and the outer invocation gets confused.

Possible fix: block xcb event handlers while waiting for an xcb response.

pipcet:
Unfortunately, I'm seeing another bug that I think is due to using the current CVS version of Emacs, but maybe this patch will make it clear to you what I think is happening and how to fix it:
https://github.com/pipcet/xelb/commit/7adf1142a602cc260e50e4c21bf094fb7076b75b

ch11ng:
I think X replies arrive sequentially so it's probably impossible for the reply of an "inner" request to come before that of an "outer" one.

I have concerns about the comparison of sequence numbers though; they are truncated to 16 bits in replies and errors and will get overflowed after some use. Although clearly this is not the cause of this issue.

pipcet:
Hmm. I've not seen the issue running with these changes:
https://github.com/ch11ng/xelb/compare/master...pipcet:master

Maybe it's the right fix with the wrong explanation.

pipcet:
(though obviously it needs more work to handle multi-reply requests, and also to use `incf` and `decf`)

pipcet:
Well, I think here's another problem, though I haven't been able to verify it yet:
`xcb:-cache-request` calls `xcb:flush` which calls `process-send-string` which calls the connection filter which calls an event handler which calls `xcb:-cache-request`. The inner `xcb:-cache-request` modifies the connection's `request-cache`, the outer `xcb:-cache-request` doesn't notice, sets its `cache` to `[]`, and drops a message to the X server.

Regardless of whether this is causing this specific bug, I really don't think we should recursively enter an XCB event handler if we're already in an XCB event handler for the same connection.

ch11ng:
I agree that we should prevent more than one event handlers from running currently. I intended to do so but failed; a simple `event-lock` is not enough.

But this issue is still worth investigating and we should take this chance. I'd suggest you to insert some `xcb:-log` forms in the process filter to show how those sequence numbers are actually modified.

pipcet:
I am still investigating this bug, and I would like to retract my earlier theories; I do not understand what's happening at all, though I believe it might be related to Emacs bug 21380: http://debbugs.gnu.org/cgi/bugreport.cgi?bug=21380.

It's quite possible that XELB is entirely innocent here, though I still think we should change some minor things to make it more resilient.

If you want to, feel free to close this PR and I'll open new ones once I'm convinced they actually help.

ch11ng:
You may open a pull request for that race condition fix though, or I will fix it myself.

ch11ng:
Can you verify if this problem still exists with ch11ng/xelb@846f4c098b5545312ad3eec6ed3d8bce3152f88b?

ch11ng:
I think this PR should be closed until someone can reproduce the problem.

PR #69, opened 2015-09-21
Add next/previous functionality for workspaces

### LINKED ISSUE #68: Could we have Next / Previous workspace functions ?
This would be really helpful to have.

CestDiego:
Also supports wrapping when there is an overflow or underflow. Closes #68


ch11ng:
See #68.

PR #99, opened 2015-11-15
Limit X host-based auth permissions.

wgreenhouse:
We don't need more than the currently logged in user
to have access to the X session, so don't grant X host access
to other users.


ch11ng:
Thanks! The file is just an example and the users are supposed to create their own `xinitrc` files to meet their needs.  Anyway, it's always  good to ensure security.

PR #108, opened 2016-02-06
Resize minibuffer only when it's in its own frame

fjl:
0e4055d3392 introduced a few calls to exwm-workspace--resize-minibuffer
in various places. This function only works when the minibuffer is
displayed in its own frame but was called unconditionally in some cases.

Fix it by wrapping all calls in an appropriate conditional and add an
assertion. Also rename the function so it is clearer that it resizes a
frame, which might prevent calling it unconditionally in the future.

There seems to be another issue with the new code which I haven't been
able to track down yet: the area of the first workspace doesn't cover the whole screen
but is slightly smaller. It's only off by a few pixels though and doesn't affect all
workspaces.

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -111,7 +111,7 @@ It's updated in several occasions, and only used by `exwm-input--set-focus'.")
             (exwm-input--set-focus exwm--id)
             ;; Adjust stacking orders
             (when exwm--floating-frame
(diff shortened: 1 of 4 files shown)


ch11ng:
Merged, thanks! As for the fullscreen issue with the first workspace, I think you can try debugging `exwm-layout--set-frame-fullscreen`.  Also, the RandR module may make a difference.

PR #142, opened 2016-03-07
Change input mode of window being clicked

medranocalvo:
When clicking a mode-line of other window to switch the EXWM input
mode, `window-buffer' does not return the window whose mode-line has
been clicked, but the current one.  This change ensures that the right
window has its input mode and mode-line updated.

* exwm-input.el (exwm-input--update-mode-line): Factor out setting
`mode-line-process'.
(exwm-input--grab-keyboard, exwm-input--release-keyboard)
(exwm-input-grab-keyboard, exwm-input-release-keyboard): Make sure
the buffer of the window being clicked has its input mode updated.

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -318,6 +318,34 @@ It's updated in several occasions, and only used by `exwm-input--set-focus'.")
                      :time xcb:Time:CurrentTime))
   (xcb:flush exwm--connection))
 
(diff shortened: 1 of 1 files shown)


ch11ng:
Thanks for contributing and apparently you've signed the paperwork.

PR #145, opened 2016-03-19
Remember last buffer

medranocalvo:
In contrast with regular Emacs buffers, EXWM buffers may only be shown in a single window.  Requesting an EXWM buffer to be displayed in some window requires the original window displaying it to have some other buffer displayed instead.  This change makes the most recent buffer displayed in that window be selected (with some sensible exceptions like when it is an EXWM buffer displayed elsewhere or it is an EXWM buffer that is being covered with another buffer).


ch11ng:
Thanks.  Using the `*scratch*` buffer as a placeholder was a really lazy solution.

PR #172, opened 2016-07-17
Dynamic workspaces

medranocalvo:
A new take on dynamic workspaces.

Support for adding and removing workspaces is in the last two commits, the rest is cleanup/preparation.

Please, test in your setups, as I don't use RandR nor floating frames.

### DIFF
--- a/exwm-config.el
+++ b/exwm-config.el
@@ -39,7 +39,7 @@
   ;; 's-w': Switch workspace
   (exwm-input-set-key (kbd "s-w") #'exwm-workspace-switch)
   ;; 's-N': Switch to certain workspace
-  (dotimes (i exwm-workspace-number)
+  (dotimes (i 10)
     (exwm-input-set-key (kbd (format "s-%d" i))
                         `(lambda () (interactive) (exwm-workspace-switch ,i))))
   ;; 's-&': Launch application
--- a/exwm-core.el
+++ b/exwm-core.el
@@ -126,7 +126,7 @@
 (declare-function exwm-layout-set-fullscreen "exwm-layout.el" (&optional id))
 (declare-function exwm-layout-toggle-mode-line "exwm-layout.el")
 (declare-function exwm-workspace-move-window "exwm-workspace.el"
-                  (index &optional id))
+                  (frame-or-index &optional id))
 
 (defvar exwm-mode-map
   (let ((map (make-sparse-keymap)))
--- a/exwm-core.el
+++ b/exwm-core.el
@@ -119,6 +119,15 @@
 ;; _MOTIF_WM_HINTS
 (defvar-local exwm--mwm-hints-decorations t)
 
+(declare-function exwm-floating-hide "exwm-floating.el")
+(declare-function exwm-floating-toggle-floating "exwm-floating.el")
+(declare-function exwm-input-release-keyboard "exwm-input.el")
+(declare-function exwm-input-send-next-key "exwm-input.el" (times))
+(declare-function exwm-layout-set-fullscreen "exwm-layout.el" (&optional id))
+(declare-function exwm-layout-toggle-mode-line "exwm-layout.el")
+(declare-function exwm-workspace-move-window "exwm-workspace.el"
+                  (index &optional id))
+
 (defvar exwm-mode-map
   (let ((map (make-sparse-keymap)))
     (define-key map "\C-c\C-f" #'exwm-layout-set-fullscreen)
--- a/exwm-floating.el
+++ b/exwm-floating.el
@@ -67,13 +67,11 @@
                                      xcb:Atom:_NET_WM_ACTION_CLOSE)))))
 
 (defvar exwm-workspace--current)
-(defvar exwm-workspace--list)
-(defvar exwm-workspace-current-index)
-(defvar exwm-workspace--switch-history-outdated)
 
 (declare-function exwm-layout--refresh "exwm-layout.el" ())
 (declare-function exwm-layout--show "exwm-layout.el" (id &optional window))
 (declare-function exwm-layout--iconic-state-p "exwm-layout.el" (&optional id))
+(declare-function exwm-workspace--minibuffer-own-frame-p "exwm-workspace.el")
 
 (defun exwm-floating--set-floating (id)
   "Make window ID floating."
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -102,6 +102,9 @@ It's updated in several occasions, and only used by `exwm-input--set-focus'.")
 
 (declare-function exwm-layout--iconic-state-p "exwm-layout.el" (&optional id))
 (declare-function exwm-layout--set-state "exwm-layout.el" (id state))
+(declare-function exwm-workspace--minibuffer-own-frame-p "exwm-workspace.el")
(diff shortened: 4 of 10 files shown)


ch11ng:
Thank you for implementing this feature.  Please leave me a few days to review and test this PR.

ch11ng:
I've finished the review & test, and this is really an exciting feature. And  I've merge your PR.  Some changes are made though, to name a few:

+ I've add `exwm-workspace-number` back, and it now sets the initial workspace number instead.  The rationale:
  - For backward compatibility (it's a public interface).
  - For forward compatibility (the method for creating new workspaces might change).
  - It makes much sense for RandR module that when you have multiple screen, you need multiple workspaces.
  - It's more accessible to users.
+ `exwm-workspace-swap-workspace` and `exwm-workspace-move-workspace` have been renamed to `exwm-workspace-swap` and `exwm-workspace-move` respectively; I think it's much clearer and should introduce no ambiguity.
+ `exwm-workspace-add` and `exwm-workspace-delete` are introduced for adding and deleting workspace, because `make-frame` and `delete-frame` do not work in all situations.  And I've add `exwm-workspace-switch` the ability to add and delete workspaces.
+ `exwm-workspace-switch` now automatically create missing workspaces.  For instance, it would create the additional 6 workspaces on receiving 9 as the index when there're only 4 workspaces currently.  Otherwise it makes little sense to bind <kbd>s-0</kbd> through <kbd>s-9</kbd> in the first place.

medranocalvo:
> I've finished the review & test, and this is really an exciting feature. And I've merge your PR.

Great!

> Some changes are made though, to name a few:
> I've add exwm-workspace-number back, and it now sets the initial workspace number instead. The rationale:
For backward compatibility (it's a public interface).
For forward compatibility (the method for creating new workspaces might change).
It makes much sense for RandR module that when you have multiple screen, you need multiple workspaces.
It's more accessible to users.

It makes sense.

> exwm-workspace-swap-workspace and exwm-workspace-move-workspace have been renamed to exwm-workspace-swap and exwm-workspace-move respectively; I think it's much clearer and should introduce no ambiguity.
exwm-workspace-add and exwm-workspace-delete are introduced for adding and deleting workspace, because make-frame and delete-frame do not work in all situations. And I've add exwm-workspace-switch the ability to add and delete workspaces.
exwm-workspace-switch now automatically create missing workspaces. For instance, it would create the additional 6 workspaces on receiving 9 as the index when there're only 4 workspaces currently. Otherwise it makes little sense to bind s-0 through s-9 in the first place.

Hmm...  I'm mildly against automatically adding/removing workspaces.  Personally, it's something I prefer not being done on my behalf, and I fear it might constrain/difficult different workspace management in the future.  It might be best left for a separate command (`exwm-workspace-switch-or-create`?), or even an `exwm-config` command (the `s-{1..9}` bindings are ultimately only a configuration suggestion).

The rest is great.

> EDIT: I forgot to mention that there still remains a problem, that when EXWM is configured to use dedicated minibuffer frame, killing the last workspace would hang Emacs.

Indeed.  Advising `delete-frame' might be the only option, as it seems there's no hook able to avoid the frame being deleted.  In my opinion this is an Emacs bug (missing feature): one should not be able to (interactively?) delete the last non-minibuffer frame.

I see you fixed quite some omissions as well, and made a lot of improvements.  Thank you very much.

ch11ng:
> Hmm... I'm mildly against automatically adding/removing workspaces. Personally, it's something I prefer not being done on my behalf, and I fear it might constrain/difficult different workspace management in the future. It might be best left for a separate command (`exwm-workspace-switch-or-create`?), or even an `exwm-config` command (the `s-{1..9`} bindings are ultimately only a configuration suggestion).

I knew it can be controversial.  I'll add a new command for that.

The <kbd>s-N</kbd> keys are supposed to be configured by users if they want to use up to `N + 1` workspaces, since setting up a new keybinding can be troublesome.  Maybe the keys should bind the new command (`exwm-workspace-switch-create`).

There's another behavior I want to alter: do not switch to the newly created workspace by default (instead make `exwm-workspace-add` do so), or there would be no way to add a workspace in the background.  This can be useful when we add many workspaces at the same time for example.  What do you think about it?

> Indeed. Advising `delete-frame' might be the only option, as it seems there's no hook able to avoid the frame being deleted. In my opinion this is an Emacs bug (missing feature): one should not be able to (interactively?) delete the last non-minibuffer frame.

I'll try to figure out what the problem really is, and see if we should fix this bug right here or in the upstream.

medranocalvo:
> I knew it can be controversial. I'll add a new command for that.
>
> The s-N keys are supposed to be configured by users if they want to use up to N + 1 workspaces, since setting up a new keybinding can be troublesome. Maybe the keys should bind the new command (exwm-workspace-switch-create).  

Thank you.  After trying it some more, I must say there's something about its magic I like, but then unexpected things happen.  For example, if I'm in a workspace with XTerm focused, then switch to another (not yet existing) workspace, XTerm moves with me (due to `make-frame` selecting the origin buffer for the new frame).  In summary, I think this command will be loved, but it's too complex to fit in a "primitive" such as `exwm-workspace-switch`.

> There's another behavior I want to alter: do not switch to the newly created workspace by default (instead make exwm-workspace-add do so), or there would be no way to add a workspace in the background. This can be useful when we add many workspaces at the same time for example. What do you think about it?

Hmm... I see your point.  I feel that `make-frame` should continue to act as in regular Emacs: create a new frame (a workspace in EXWM) and focus it.  That way one can use, for example, `find-file-other-frame` to open some file in a new workspace and switch to it.  I'd say _window-managery_ behaviours should be implemented in EXWM commands; `exwm-workspace-add` could accept a parameter (or observe some variable) to add the workspace on the background.  Otherwise we could improve core's `make-frame` to accept such an argument or make sure the regular way to create a frame in the background works in EXWM.  The latter has kind of a multiplicative effect, as all packages creating frames end up integrating perfectly with EXWM, instead of needing to adapt to its peculiarities.

This (kind of) creates a frame in the background in regular Emacs:
```{.elisp}
(let ((f (selected-frame)))
  (make-frame)
  (select-frame-set-input-focus f))
```

ch11ng:
> Hmm... I see your point. I feel that `make-frame` should continue to act as in regular Emacs: create a new frame (a workspace in EXWM) and focus it. That way one can use, for example, `find-file-other-frame` to open some file in a new workspace and switch to it. I'd say *window-managery* behaviours should be implemented in EXWM commands; `exwm-workspace-add` could accept a parameter (or observe some variable) to add the workspace on the background. Otherwise we could improve core's `make-frame` to accept such an argument or make sure the regular way to create a frame in the background works in EXWM. The latter has kind of a multiplicative effect, as all packages creating frames end up integrating perfectly with EXWM, instead of needing to adapt to its peculiarities.

Yes, it makes more sense.  I realized it's easy to create workspaces in the background, so I'll make the two schemes coexist and stick with the current behavior.

ch11ng:
It turns out Emacs doesn't freeze after deleting the last workspace (it just appears to be so).  I've fixed it in 6ecd8b921fade26f30bdf4c0b047fcc9d9b57e9d by creating a new workspace.

medranocalvo:
Thank you.  By the way, latest changes are great.

ch11ng:
And thank you for your contribution!

PR #176, opened 2016-07-30
set fullscreen-frame-count to 0 on exiting

uemurax:
I use Emacs as a daemon and execute `emacsclient -c` at the end of `.xsession`.
First time, everything is good.
However, if I exit the X session using `save-buffers-kill-terminal` and restart X,
then an error occurs and the X session goes down.
Enabling debug, I found this is because
in the function `exwm-workspace--post-init`, the variable `exwm-workspace--fullscreen-frame-count`
is used as a numeric value, but at the end of the function, the variable is set to `nil`
so that at the second call of the function, `wrong-type-argument` is signaled.
To fix the bug, I propose to reset `exwm-workspace--fullscreen-frame-count` to `0`
in the function `exwm-workspace--exit`.

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -1401,6 +1401,7 @@ applied to all subsequently created X frames."
         exwm-workspace--list nil
         exwm-workspace--client nil
         exwm-workspace--minibuffer nil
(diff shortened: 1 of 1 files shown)


ch11ng:
Looks good; merged.

PR #200, opened 2016-09-30
Use read-buffer-to-switch in exwm-workspace-switch-to-buffer

fjl:
read-buffer-to-switch elides the current buffer from the list of completions and selects a sane default.

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -783,7 +783,7 @@ INDEX must not exceed the current number of workspaces."
              (rename-buffer (substring (buffer-name) 1))))))
      (prog1
          (with-local-quit
-           (list (get-buffer (read-buffer "Switch to buffer: " nil t))))
+           (list (get-buffer (read-buffer-to-switch "Switch to buffer: "))))
        ;; Hide buffers on other workspaces
        (unless exwm-workspace-show-all-buffers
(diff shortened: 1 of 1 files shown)


fjl:
Quick question: What is the purpose of `exwm-workspace-switch-to-buffer` anyway?

I'm asking because I got confused by its existence. What I wanted to achieve is that switching to an exwm-mode buffer moves the X window to the current workspace if necessary. I tried rebinding <kbd>C-x b</kbd> to `exwm-workspace-switch-to-buffer` for a while and found that it breaks ivy's virtual buffers and uses the wrong read function (fixed in this PR).

Some further investigation shows that I can just set `exwm-workspace-show-all-buffers` and `exwm-layout-show-all-buffers` to `t` and then regular `switch-to-buffer` and all its derivatives just work. Maybe this is a documentation issue.

ch11ng:
Thanks! I didn't notice the problem since I always use `ido-everywhere`.

`exwm-workspace-switch-to-buffer` is intended to be used when those two variables you mentioned are not set, that is, `exwm-mode` buffers are not shared among workspaces.

> It would make sense to have it if it actually switched to the workspace that the X window is on, but right now it pulls the window into the current workspace.

This is not implemented yet, but it's possible.

PR #201, opened 2016-10-02
make exwm-workspace-switch-to-buffer switch workspaces

fjl:
As per discussion in #200, I'd like to propose that `exwm-workspace-switch-to-buffer` should switch workspaces.

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -772,10 +772,10 @@ INDEX must not exceed the current number of workspaces."
 
 ;;;###autoload
 (defun exwm-workspace-switch-to-buffer (buffer-or-name)
-  "Make the current Emacs window display another buffer."
+  "Switch to a workspace displaying the given buffer."
   (interactive
    (let ((inhibit-quit t))
-     ;; Show all buffers
+     ;; Show all buffers temporarily.
      (unless exwm-workspace-show-all-buffers
        (dolist (pair exwm--id-buffer-alist)
          (with-current-buffer (cdr pair)
@@ -784,7 +784,7 @@ INDEX must not exceed the current number of workspaces."
      (prog1
          (with-local-quit
            (list (get-buffer (read-buffer-to-switch "Switch to buffer: "))))
-       ;; Hide buffers on other workspaces
+       ;; Hide buffers on other workspaces again.
        (unless exwm-workspace-show-all-buffers
          (dolist (pair exwm--id-buffer-alist)
            (with-current-buffer (cdr pair)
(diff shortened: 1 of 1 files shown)


ch11ng:
Thanks! I've just tried this and it seems working for me.  But I think it'd be better if you can make this a new command or some users may get confused by the change (and `exwm-workspace-switch-to-buffer` is naturally analogous to `switch-to-buffer`).

fjl:
Sorry didn't have much time last few weeks.

What should it be renamed to? I thought `exwm-workspace-switch-to-buffer` made sense because of the analogy to `exwm-workspace-switch`. The best I can come up with is `exwm-workspace-switch-by-buffer` but then they're too similar...

ch11ng:
Perhaps simply `exwm-workspace-switch-to-buffer-1`?

medranocalvo:
I'd say `exwm-workspace-select-buffer`, in analogy with `select-*` functions, particularly the interactive command `select-frame-by-name`.

Best regards,
Adrián.

madnificent:
Is there interest in reviving this?

timor:
> Is there interest in reviving this?

I would be interested.  Common use case for me: `S-f` switches to web browser buffer on my setup.  Most of the time, I don't want to pull the browser into my workspace if it visible on another one.  Since I do that quite often, it would save quite a bit of key strokes.

2ndemosthenes:
@medranocalvo Any plans on merging this?
It's quite annoying.

medranocalvo:
@2ndemosthenes, thank you for bringing this up.

I think it is a fine use case.

I don't think this should be implemented in EXWM (necessarily, see below).

In my view, a key feature of EXWM is that it lets us treat X windows as Emacs' buffers, thus apply Emacs' window management to X windows.  With that I mean not only Emacs' default window management (`other-window`, etc.), but also `evil` or `avy` or `shackle`... To each their own.

The feature proposed here is the ability to select the window displaying a buffer in another frame instead of displaying the buffer in current window (where the buffer is an `exwm-mode` buffer, i.e., an X window).

Does Emacs already offer a way to do this?

If it does, we should write about it in the Cookbook, or showcase it in `exwm-config-example`.  Maybe we should offer a function in order to ease configuration or discoverability, or maybe we should tailor the function to EXWM (for example, search only in EXWM workspaces).

If Emacs doesn't: it should!  We should patch Emacs to support this.

---
*...checks Emacs...*

It seems to me that the following is Emacs' way of doing this:

```
(pop-to-buffer
 (read-buffer "Select buffer: " (other-buffer))
 '((display-buffer-reuse-window) . ((reusable-frames . visible))))
```

See also `display-buffer-alist`.

---

2ndemosthenes:
Thanks you.

medranocalvo:
@2ndemosthenes (and everyone else):  I would appreciate your thoughts on the following:

> If it does, we should write about it in the Cookbook, or showcase it in exwm-config-example. Maybe we should offer a function in order to ease configuration or discoverability, or maybe we should tailor the function to EXWM (for example, search only in EXWM workspaces).

2ndemosthenes:
@medranocalvo
Yes, I think it would be very helpful to add it Cookbook or the config since many may want to do this but dont know how (or maybe think it can't be done)

cherryramatisdev:
Do we have any workarounds to get this functionality right now? I would love to select an EXWM buffer and navigate to it's workspace properly

PR #223, opened 2017-02-22
Use let instead of setq when setting exwm-input--line-mode-passthrough

Stebalien:
That way, exwm won't clobber user modifications to this variable.


ch11ng:
Thanks! I've also exposed that interface.

PR #311, opened 2017-10-13
avoid running the focus timer immediately

### LINKED ISSUE #310: The update focus timer introduces a significant input lag
To test, open a scratch buffer, repeatedly type the same character (e.g., 400 x 'a'), then use the arrow keys to move the cursor back and forth over the 'a's. With `exwm-input--update-focus-interval` set to the default value (0.01) the cursor stutters as it's moved over the text. With this value set to, e.g., 1, the cursor moves smoothly. This is a contrived example but there has to be a better way to do this than to poll every 0.01 seconds (i.e., use a separate process/thread?).

Stebalien:
see https://www.gnu.org/software/emacs/manual/html_node/elisp/Idle-Timers.html

fixes #310

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -189,7 +189,8 @@ This value should always be overwritten.")
     (when exwm-input--update-focus-timer
       (cancel-timer exwm-input--update-focus-timer))
     (setq exwm-input--update-focus-timer
-          (run-with-idle-timer exwm-input--update-focus-interval nil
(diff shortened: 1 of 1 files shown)


Stebalien:
It would still be better to avoid a timer but this makes emacs lag significantly less.

Stebalien:
Never mind. This fixes it because it effectively set a really long timer when `(current-idle-time)` returns `nil`.

PR #334, opened 2017-11-25
Allow customizing default output

hying-caritas:
* exwm-randr.el (exwm-rand-default-output): New variable for
customizing default output.
(exwm-randr--refresh): Use it


ch11ng:
Actually you can specify a primary output by passing the `--primary` option to xrandr, and EXWM would pick that output as the default one.

PR #378, opened 2018-02-28
Fix EXWM buffers not being hidden on workspace switch

bryal:
When I switch workspaces, EXWM buffers are not correctly hidden, but stays visible and in the way of the buffers I actually want to use. I tracked it down to the `output-{new,old}` objects being equal component wise, but not reference wise in this `cond` clause. Using the less-strict component-wise equality test of `equal` fixed it for me.

I see there are many uses of `eq` in the EXWM source in general. Maybe there are more bugs in here due to the use of too-strict equality tests with `eq`. Just a thing to consider.

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -544,8 +544,8 @@ for internal use only."
          (t
           (dolist (w exwm-workspace--list)
             (when (and (exwm-workspace--active-p w)
-                       (eq output-new
-                           (frame-parameter w 'exwm-randr-output)))
+                       (equal output-new
+                              (frame-parameter w 'exwm-randr-output)))
               (exwm-workspace--set-active w nil)
               (setq workspaces-to-hide (append workspaces-to-hide (list w)))))
           (exwm-workspace--set-active frame t)))
@@ -756,8 +756,8 @@ INDEX must not exceed the current number of workspaces."
           ;; Floating.
           (setq container (frame-parameter exwm--floating-frame
                                            'exwm-container))
-          (unless (eq (frame-parameter old-frame 'exwm-randr-output)
-                      (frame-parameter frame 'exwm-randr-output))
+          (unless (equal (frame-parameter old-frame 'exwm-randr-output)
+                         (frame-parameter frame 'exwm-randr-output))
             (with-slots (x y)
                 (xcb:+request-unchecked+reply exwm--connection
                     (make-instance 'xcb:GetGeometry


medranocalvo:
Thank you for tracking and fixing this.

I suggest, instead of changing the predicate to `equal`, to define a new function `exwm-randr--output-equal-p` that takes care of comparing two `'exwm-randr-output` frame parameters.  That way other modules don't need to know what the content of this frame parameter is, or how should it be compared.

The following places must be changed as well:
* https://github.com/ch11ng/exwm/blob/master/exwm-workspace.el#L547
* https://github.com/ch11ng/exwm/blob/master/exwm-workspace.el#L759

I had a brief look at other uses of `eq`, but they seem all right: they compare frames, windows, buffers, symbols, and X atoms and properties.

bryal:
I'm not very good with elisp and the emacs API in general, and I'm not sure I quite understand what you mean. 

Do you mean that `exwm-randr--output-equal-p` should be defined as `equal`, so that dependants on the `exwm` library know to use that equivalence relation function, instead of needing to find out by themselves whether to use `eq` or `equal` when comparing these frame parameters?

And just because I'm curious, do you know why the frame parameters are the only objects that require `equal` instead of `eq`?

medranocalvo:
> Do you mean that exwm-randr--output-equal-p should be defined as equal, so that dependants on the exwm library know to use that equivalence relation function, instead of needing to find out by themselves whether to use eq or equal when comparing these frame parameters?

Exactly. Define `exwm-randr--output-equal-p` as `equal` and use that instead of springing `equal`s all over the place.  The intention is to help us, as the code will be clearer, less complicated to change and evolve.

Nevertheless, please wait on input from @ch11ng.  I'm just another user.

> And just because I'm curious, do you know why the frame parameters are the only objects that require equal instead of eq?

The reason is that `eq` returns `t` when its two arguments are one and the same value, not when they are two values with the same content. That depends on implementation details, but at least they are documented: https://www.gnu.org/software/emacs/manual/html_node/elisp/Equality-Predicates.html. 

> I'm not very good with elisp and the emacs API in general, and I'm not sure I quite understand what you mean.

Don't hesitate to ask if you have any doubt.  I can't promise to have an answer, though :-).

ch11ng:
Replacing `eq` with `equal` would be fine for now.  Please fix the other two places so I can merge it.

bryal:
Changed `eq` to `equal` in the two places pointed out by medranocalvo.

PR #381, opened 2018-03-01
Support disabling default `exwm-mode-map' bindings

medranocalvo:
This lets `(define-key exwm-mode-map (kbd "C-c") nil)` recover <key>C-c</key> for XTerm.

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -566,7 +566,7 @@ instead."
                      ;;
                      (memq event exwm-input--global-prefix-keys)
                      (memq event exwm-input-prefix-keys)
-                     (assq event (cdr exwm-mode-map))
+                     (lookup-key exwm-mode-map (vector event))
                      (gethash event exwm-input--simulation-keys)))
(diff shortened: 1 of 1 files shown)


exwm-input.el
```diff
@@ -566,7 +566,7 @@ instead."
                      ;;
                      (memq event exwm-input--global-prefix-keys)
                      (memq event exwm-input-prefix-keys)
-                     (assq event (cdr exwm-mode-map))
+                     (lookup-key exwm-mode-map [event])
```
ch11ng: Should be `(lookup-key exwm-mode-map (vector event))`

medranocalvo:
Ooops; it *definitely* worked for **disabling** :-/.  Amended, thanks for the catch.

ch11ng:
I've merged a slightly different one (dd0f10775a1893abe3b5c753e9db12d1397c2f6b).

medranocalvo:
Oh, that's better.  Thank you.

PR #467, opened 2018-08-13
Exwm log buffer

medranocalvo:
Using `message' to log debugging information is cumbersome, as the output
appears constantly in the minibuffer, obscuring prompts and other information.
In the case of long messages, it might resize the minibuffer, which causes EXWM
to perform additional actions due to the log output.

This change reimplements EXWM debug logging using a separate
buffer (*EXWM-DEBUG*).  Basic functionality, like scrolling when point is at the
end of the buffer is maintained.

I've been using this for some time.  Let me know whether you find it interesting or whether you can think of any improvements.

### DIFF
--- a/exwm-core.el
+++ b/exwm-core.el
@@ -31,6 +31,7 @@
 (require 'xcb)
 (require 'xcb-icccm)
 (require 'xcb-ewmh)
+(require 'exwm-debug)
 
 (eval-and-compile
   (defvar exwm-debug-on nil "Non-nil to turn on debug for EXWM."))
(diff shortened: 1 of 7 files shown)


ch11ng:
Thanks! The issue has long been bothering me.

ch11ng:
@medranocalvo The more I use it the more I find this feature extremely useful! Could you please move it to XELB so both libs can use it? Only some renaming work is required I think.  A few suggestions BTW:

* Make the logging macros (xcb:-log and exwm--log) functions.  This allows us to toggle the switch at runtime.  I originally thought no-op function calls can introduce unnecessarily overheads but it turns out to be not true.
* Add a command to toggle the switch.

medranocalvo:
Happy to hear that.  Please, see ch11ng/xelb#22 and #478.

ch11ng:
@medranocalvo Thanks! I've had them merged.

PR #473, opened 2018-08-28
Bugfix for cursor warping

WJCFerguson:
When I move to a workspace to the right, the warping doesn't occur because the `win-x` is negative.  This seems to fix the issue.

Also explicitly specifies the frame for `(frame-pixel-height)`, though `frame` is `(selected-frame)` anyway, so this appears to have no functional consequences.

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -612,7 +612,7 @@ for internal use only."
         (when (or (< win-x 0)
                   (< win-y 0)
                   (> win-x (frame-pixel-width frame))
-                  (> win-y (frame-pixel-height)))
+                  (> win-y (frame-pixel-height frame)))
           (xcb:+request exwm--connection
               (make-instance 'xcb:WarpPointer
                              :src-window xcb:Window:None
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -609,7 +609,9 @@ for internal use only."
               (make-instance 'xcb:QueryPointer
                              :window (frame-parameter frame
                                                       'exwm-outer-id)))
-        (when (or (> win-x (frame-pixel-width frame))
+        (when (or (< win-x 0)
+                  (< win-y 0)
+                  (> win-x (frame-pixel-width frame))
                   (> win-y (frame-pixel-height)))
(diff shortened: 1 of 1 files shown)


WJCFerguson:
I've also been messing with having the cursor warp to the current window, rather than the frame.  I do prefer window-warping, but it occurs to me that a superior solution might be to remember the cursor position per workspace and return to that.

Before this exwm built-in warping was working, I had some custom config (calling the `swarp` external program) that would warp the cursor if you selected the already-current workspace, i.e. when I pressed the workspace hotkey a second time.  I quite like this behavior, and could code it up for a PR, but managing the possible variations of this and/or remembering positions via customization starts to get over-complex.

Thoughts?

medranocalvo:
Hello @WJCFerguson, thank you for the patch!  The fix seems sensible to me, and the explicit `frame` is very welcome, as it is not always clear whether the selected frame is the one we are acting on.

I won't merge this yet as I'm still getting my feet wet with the project, and don't have much time on my hands.  I'll probably be able to do it during the weekend, if @ch11ng does not do it before.

With regards to window warping, `exwm-workspace-warp-cursor` could be made to be either `nil`, `warp-to-frame`, `warp-to-window` or `warp-to-last`.  It would define how to warp the cursor upon workspace focus change.  I think that workspaces remember the previously selected window via the `exwm-selected-window` frame parameter, perhaps that can be used.  `warp-to-last` would be a great default behavior (i.e., for `(setq warp-to-selected-window t)` ), but seems the most messy to implement.

ch11ng:
@WJCFerguson thanks! The fixes look good to me.  I'm merging them.

@medranocalvo not sure if we should implement something like `warp-to-selected-window` but we could provide a simplified one.  Basically `warp-to-last` would require querying and recording the position of pointer when switching from a workspace (possibly by introducing another frame parameter like `exwm-selected-window`), restoring and resetting it when switching to that workspace from another monitor later and resetting it in other corner cases.  Both `warp-to-window` and `warp-to-last` fall back to `warp-to-workspace` whenever necessary.

PR #476, opened 2018-08-30
Restack X-Windows after being mapped to ensure EnterNotify events fire

### LINKED ISSUE #422: Switching to other workspaces while hovering an X window fails when focus follows mouse is enabled
Switching to other workspaces fails when _focus follows mouse_ is enabled `(setq mouse-autoselect-window t)` when the mouse pointer is hovering an X window.

# Recipe
1. Start EXWM
```
emacs -Q -L /path/to/exwm/ -L /path/to/xelb/ \
        --eval '(setq mouse-autoselect-window t)' \
        --eval '(setq focus-follows-mouse t)' \
        --eval '(require 'exwm)' \
        --eval '(exwm-init)'
```
2. Start XTerm, delete all other emacs windows in the frame.
3. Create a second EXWM workspace, switch to it.
4. Switch back to the first workspace. Before switching, make sure the mouse pointer stays around the center of the buffer, so that upon switching to the first workspace it will be over the XTerm window.
5. Without moving the mouse out of the XTerm window, nor clicking anywhere, switch to the 2nd workspace.  The workspace switch reverts back to the first workspace.

Let me know if the description is not clear.

# Workaround
Clicking on the X window before switching to other workspace lets the switch happen.

medranocalvo:
```
* exwm-manage.el (exwm-manage--on-MapNotify, exwm-manage--init):
Restack X windows after being mapped in order to ensure they receive
an EnterNotify event when running under XQuartz.
```

This change fixes #422 for me.  Let me know of any comments.

### DIFF
--- a/exwm-manage.el
+++ b/exwm-manage.el
@@ -665,6 +665,22 @@ border-width: %d; sibling: #x%x; stack-mode: %d"
       (exwm--log "id=#x%x" window)
       (exwm-manage--unmanage-window window t))))
 
+(defun exwm-manage--on-MapNotify (data _synthetic)
+  "Handle MapNotify event."
+  (let ((obj (make-instance 'xcb:MapNotify)))
+    (xcb:unmarshal obj data)
+    (with-slots (window) obj
+      (when (assoc window exwm--id-buffer-alist)
+        (exwm--log "id=#x%x" window)
+        ;; With this we ensure that a "window hierarchy change" happens after
+        ;; mapping the window, as some servers (XQuartz) do not generate it.
+        (xcb:+request exwm--connection
+            (make-instance 'xcb:ConfigureWindow
+                           :window window
+                           :value-mask xcb:ConfigWindow:StackMode
+                           :stack-mode xcb:StackMode:Above))
+        (xcb:flush exwm--connection)))))
+
 (defun exwm-manage--on-DestroyNotify (data synthetic)
   "Handle DestroyNotify event."
   (unless synthetic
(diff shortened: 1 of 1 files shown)


exwm-manage.el
```diff
@@ -665,6 +665,20 @@ border-width: %d; sibling: #x%x; stack-mode: %d"
       (exwm--log "id=#x%x" window)
       (exwm-manage--unmanage-window window t))))
 
+(defun exwm-manage--on-MapNotify (data _synthetic)
+  (let ((obj (make-instance 'xcb:UnmapNotify)))
```
ch11ng: You actually mean `xcb:MapNotify`.

exwm-manage.el
```diff
@@ -665,6 +665,20 @@ border-width: %d; sibling: #x%x; stack-mode: %d"
...
+    (xcb:unmarshal obj data)
+    (with-slots (window) obj
+      (exwm--log "id=#x%x" window)
+      ;; With this we ensure that a "window hierarchy change" happens after
+      ;; mapping the window, as some servers (XQuartz) do not generate it.
+      (xcb:+request exwm--connection
+          (make-instance 'xcb:ConfigureWindow
+                         :window window
+                         :value-mask xcb:ConfigWindow:StackMode
+                         :stack-mode xcb:StackMode:Above))
```
ch11ng: We may receive MapNotify from other sources.  So perhaps we should test whether the X window is in `exwm--id-buffer-alist`?

exwm-manage.el
```diff
@@ -692,6 +706,8 @@ border-width: %d; sibling: #x%x; stack-mode: %d"
               #'exwm-manage--on-ConfigureRequest)
   (xcb:+event exwm--connection 'xcb:MapRequest #'exwm-manage--on-MapRequest)
   (xcb:+event exwm--connection 'xcb:UnmapNotify #'exwm-manage--on-UnmapNotify)
+  (when (eq system-type 'darwin)
```
ch11ng: Please consider applying this 'workaround' to all platforms if it has no side-effect.  It would help reducing the testing scenarios.

medranocalvo:
> You actually mean xcb:MapNotify.

Corrected.

> We may receive MapNotify from other sources. So perhaps we should test whether the X window is in exwm--id-buffer-alist?

Corrected.

> Please consider applying this 'workaround' to all platforms if it has no side-effect. It would help reducing the testing scenarios.

It now applies to all platforms.  I can't foresee any bad effect except the useless restacking.  Please, test it in your environment.

Thank you for the review.

Issue #359, opened 2018-02-04
When  exwm-workspace-minibuffer-postion is set to bottom active process in term steals focus from ivy


Sbozzolo:
I reported this issue to the ivy project [here](https://github.com/abo-abo/swiper/issues/1429), but it turned out to be related do EXWM.
Essentially, when  `(setq exwm-workspace-minibuffer-position 'bottom)` is in my configuration files and I am running a process in a term-mode, `ivy` loses focus and it is impossible to move the cursor back to the minibuffer.

![gif](https://user-images.githubusercontent.com/9167485/35782325-5e306966-09b3-11e8-8bc4-06ccece3a223.gif)

dakra:
I have maybe a related problem.

I have some global hydras to launch apps or toggle settings
and they work fine in Emacs buffers but in exwm buffers (mainly pidgin, firefox)
the hydra opens but then the focus is in the exwm buffer and no way to select
the minibuffer again.

It didn't bother me too much to investigate and report as I do almost everything in Emacs
buffers and thought it has to do with mouse over focus which apparently nobody
besides me uses.
But seeing your issue I think it could be the same problem.

ch11ng:
Are you on Emacs 25? It seems to be a problem with pre-26 releases.

Follpvosten:
I'm having a similar issue, but it's much worse - whenever I press tab to invoke completions in the minibuffer, the focus changes to the current workspace buffer (and all of the other problems mentioned here apply), while the completion buffer stays open. I can use the mouse to click options in the completion buffer, but that's the only way to make it work as intended. If I instead invoke `other-window` a couple of times, the completion window vanishes and the echo area shows "Quit". That basically makes EXWM unusable when the minibuffer/echo area is hidden. I hope Emacs 26 is gonna fix that (as i'm currently on 25).

dakra:
I'm on Emacs 27/current git master with latest Exwm and the problem is still
there.

Easy way to reproduce is to install [hydra](https://github.com/abo-abo/hydra)
and then create a simple hydra like:

``` emacs-lisp
(defhydra hydra-exwmtest (:color blue)
  "exwm test"
  ("m" (lambda ()
         (interactive)
         (message "Doesn't work."))))
```

Then running `M-x hydra-exwmtest/body` shows the 'm' hint
and in a Emacs buffer you can press 'm' to show the message but in a X11 buffer
you can not.

ch11ng:
[...]

PR #484, opened 2018-09-14
Fix focus jumps with minibuffer-only frames

medranocalvo:
* exwm-input.el (exwm-input--update-focus): Focus the minibuffer's
frame, which is different than that of the
`minibuffer-selected-window' on minibuffer-only frames.
(exwm-input--on-minibuffer-setup): Remove unneeded function.

@ch11ng, I couldn't find a use for `exwm-input--on-minibuffer-setup`.  Am I missing something?

Should help with #359.


medranocalvo:
Updated the proposal removing the hook on `exwm-input--exit`.  Thank you for your review.

Issue #250, opened 2017-04-23
Add an option to hide mode-line for floating windows


QiangF:
Some applications like gimp and ug NX create a lot of small windows, and each one showing a mode-line is not pretty. What about add a function to toggle mode-line of all floating windows?

ch11ng:
Is putting `exwm-layout-hide-mode-line` into `exwm-floating-setup-hook` and `exwm-layout-show-mode-line` into `exwm-floating-exit-hook` enough for this?

QiangF:
I have create a function template, it is not working now.

    (defun my-toggle-exwm-floating-modeline ()
        (interactive)
        (if (member 'exwm-layout-hide-mode-line 'exwm-floating-setup-hook)
        (progn
            (if current-buffer-is-floating
                (exwm-layout-hide-mode-line))
            (remove-hook 'exwm-floating-setup-hook 'exwm-layout-hide-mode-line)
            (remove-hook 'exwm-floating-exit-hook 'exwm-layout-show-mode-line))
        (add-hook 'exwm-floating-setup-hook 'exwm-layout-hide-mode-line)
        (add-hook 'exwm-floating-exit-hook 'exwm-layout-show-mode-line)
        (if current-buffer-is-floating
            (exwm-layout-show-mode-line))))

QiangF:
The modeline is noly useful to show window title and guide key. For people likes mouse clicking, how to set middle mouse click to close the buffer that is clicked by force without asking question?

ch11ng:
> I have create a function template, it is not working now.

I think you only need these `add-hook` lines and should use `exwm-layout-toggle-mode-line` to toggle the mode-line.

> how to set middle mouse click to close the buffer that is clicked by force without asking question and Ctrl + middle click to hide floating window?

For the first question, you'll need to let-bind `kill-buffer-query-functions` to nil.  For the second one, just call `exwm-floating-hide`.

QiangF:
Do we need two more mouse down event?

    (defvar exwm-input-close-event 's-down-mouse-2
    "Emacs event to close a window.")

    (defvar exwm-input-hide-floating-event 'c-s-down-mouse-2
    "Emacs event to hide a floating window.")

ch11ng:
Instead we can make `exwm-input-set-key` accept mouse events.

medranocalvo:
You can already close a floating window with the mouse by clicking on the "EXWM" menu on the mode-line and clicking on "Close window".

[...]

PR #485, opened 2018-09-14
Handle mouse events in line mode

medranocalvo:
* exwm-input.el (exwm-input--forward-event-to-emacs-p): Predicate
checking whether an event should be forwarded to Emacs.
(exwm-input--on-KeyPress-line-mode): Use it.
* exwm-input.el (exwm-input--cache-event): Protect against
translations swallow the event.
* exwm-input.el (exwm-input--on-ButtonPress): Forward bound mouse
events to Emacs when in line-mode.

Could help with #250.

@ch11ng, do you think this is a good approach?

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -87,7 +87,7 @@ defined in `exwm-mode-map' here."
                        value))))
 
 (defcustom exwm-input-line-mode-passthrough nil
-  "Non-nil makes 'line-mode' forwards all events to Emacs."
+  "Non-nil makes 'line-mode' forward all events to Emacs."
   :type 'boolean)
 
 ;; Input focus update requests should be accumulated for a short time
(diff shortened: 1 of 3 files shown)


exwm-input.el
```diff
@@ -414,11 +415,21 @@ ARGS are additional arguments to CALLBACK."
...
+               (if (and (derived-mode-p 'exwm-mode)
+                          (eq exwm--on-KeyPress
+                              'exwm-input--on-KeyPress-line-mode)
+                          (setq read-event
+                                (exwm-input--mimic-read-event button-event))
+                          (exwm-input--forward-event-to-emacs-p read-event))
+                   (progn
+                     ;; The event should be forwarded to emacs
+                     (exwm-input--cache-event read-event)
+                     (exwm-input--unread-event button-event))
```
ch11ng: `exwm-input--unread-event` should be defined before this so it would cause a compile warning.

exwm-input.el
```diff
@@ -414,11 +415,21 @@ ARGS are additional arguments to CALLBACK."
...
                    (select-window window)
-                 (setq window
-                       (get-buffer-window (exwm--id->buffer event) t))
+                 (setq window (get-buffer-window buffer t))
                  (when window (select-window window))))
-             ;; The event should be replayed
-             (setq mode xcb:Allow:ReplayPointer))))
+             (with-current-buffer buffer
+               (if (and (derived-mode-p 'exwm-mode)
+                          (eq exwm--on-KeyPress
```
ch11ng: This is not aligned.

ch11ng:
It'd good if this can be made to work with `char-mode`.  Anyway this is a good start.

medranocalvo:
Resolved the issues you raised.  This new proposal reworks the input mode code in a first commit, which makes it easier to integrate the button handling code.  I've not started looking into support for mouse buttons in char-mode, but it might be O.K. to merge this anyway. Let me know what you think.

PR #541, opened 2019-02-02
Cleanup simulation key config issues

ieure:
- Rearrange defcustom for `exwm-input-simulation-keys`:
   - Original key only has one option, so probably shouldn’t be a
     `choice` type.
   - Move the "User-defined" key value to the top, since that’s the
     one someone is most likely to want.
   - Set a default value for the variable.
 - Remove the setting of `exwm-input-simulation-keys` from
   `exwm-config-default`.  When this value is set, it discards the
    simulation keys the user set up in customize.

### DIFF
--- a/exwm-config.el
+++ b/exwm-config.el
@@ -31,7 +31,8 @@
 (defun exwm-config-default ()
   "Default configuration of EXWM."
   ;; Set the initial workspace number.
-  (setq exwm-workspace-number 4)
+  (unless (get 'exwm-workspace-number 'saved-value)
+    (setq exwm-workspace-number 4)
   ;; Make class name the buffer name
   (add-hook 'exwm-update-class-hook
             (lambda ()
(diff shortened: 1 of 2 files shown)


ch11ng:
I agree with the change to the Customize interface, but I'm afraid updating the default value of `exwm-input-simulation-keys` can cause problems to users leaving this option unset before.  `exwm-config-default` was created as a handy function to help users try out this features.  With the Customize interface it can also be used to populate reasonable initial values for various user options, but it should be removed afterwards.

ieure:
Unfortunately, I don't think there's an option that doesn't break some workflow or other, so it's a matter of choosing which one to break.  I took the approach I did because I suspect that most EXWM users are following the instructions, which tell you to set up something which does not work.

The case it breaks is when someone who hasn't used custom to set `exwm-input-simulation-keys`, and isn't using `exwm-config-default`, and also isn't setting simulation keys some other way.  It seems like a corner case which is unlikely to affect many users.

ieure:
The other option that could work is to leave the default value `nil`, but also delete the `(setq exwm-input-simulation-keys ...)` from `exwm-config-default`.  However, I think that would break more users' configs, because it's a change to the *default* config, vs. breaking a case where you have a configuration so customized that you don't use `exwm-config-default` at all.

I think most EXWM users use `exwm-config-default`, because that's what the instructions say, and it has such basic functionality (like workspace switching, resetting exwm, naming buffers) that you get a broken experience without it.  This gives it the air of "this is the stuff you always need, but you can add more," not "here's an example, you should change it."

ieure:
I think `exwm-config-default` already occurs after customizations are loaded, at least it does on my setup.

After poking around a bit, I think I have a viable approach.  It also fixes the same problem with `exwm-workspace-number`, which I just noticed.  When customizations are loaded, the symbol's `saved-value` property is set to `(list value)`.  So if it has a saved custom value of `nil`, you get `(nil)`, but if there's no saved value, it's `nil`.  So I think that will work okay, though I continue to think that the user should be the only one to set a usre-customizable value.

I pushed a commit, let me know what you think.

ch11ng:
> I think `exwm-config-default` already occurs after customizations are loaded, at least it does on my setup.

You've probably put it after the auto-generated `custom-set-variables` block, which is usually at end of user init file.

You solution works anyway, so I've merged it as ec108a61dc38f84a3b290f022929331e97784842.

Issue #255, opened 2017-05-01
Killing a non-responsive GUI application results in a mini buffer question


map7:
At the moment when I try and kill a non-responsive GUI application within exwm I get the following question in the mini-buffer

Please answer y or n. 'Vlc' is not responding. Would you like to kill it? (y or n).

The problem is I cannot answer 'y' or 'n' until the frozen GUI app has been killed and I cannot kill it as I cannot get access to any buffer in exwm at this time or answer the question to kill it. The only way I fix this problem is to got to VT1 and kill all the GUI's in that workspace until I'm on an emacs buffer and then I can answer the question and continue on.

ch11ng:
Some similar problems had be reported before but we didn't have reliable way to reproduce them.  How can we reproduce this one?

map7:
I've been able to reproduce this error

Within exwm do the following
1. Start gtk_unresponsive program (s-&)
2. Click on 'Hello' button
3. Within a few seconds try and kill the buffer with C-x k
4. Answer the first question in the mini buffer
5. Now you should have the second question in which it's impossible to answer and you have to go to a virtual terminal to kill the original GUI application.

A pre-compiled version is included and also the source code. This is just the GTK gui example with a sleep in the correct place to re-enact this issue.
[exwm_unresponse_problem.tar.gz](https://github.com/ch11ng/exwm/files/1000077/exwm_unresponse_problem.tar.gz)

ch11ng:
Thanks! Your tool revealed a problem, but it does not seems to be the one your reported.  Here is my discovery:

1. If I run this tool as root, then after Step 4 I would get a empty `exwm-mode` buffer (the X window is unmapped but the application is not killed).
2. If I run this tool normally it would get killed every time after Step 4.

Did I miss something?

map7:
You must click the hello button and then quickly try and kill it. I run it as a normal user. I just tried it again and it failed.

ch11ng:
No idea but it seems I can kill it every time no matter how quickly I try.  Basically for this application EXWM would send it a SIGKILL signal followed by a KillClient X request sent in a timer.  According to your description the SIGKILL signal probably didn't take effect, and the KillClient X request was perhaps not sent at all.  I wonder if Emacs was busy in step 5.  If so, can you interrupt it with SIGUSR2?

map7:
Would I send the SIGUSR2 to the emacs process by doing the following?
$ kill -SIGUSR2 <pid of emacs>

[...]

PR #578, opened 2019-05-19
Surely grab keyboard when answering about killing unresponsive app #255

demokritos:
When an app is unresponsive, killing the buffer causes yes-or-no question in
minibuffer. But if the buffer is in char mode, we cannot answer the question.
Changing mode to line mode is also blocked. This is probably because y-or-n-p
catches the mode-transition key. So, first change mode to line mode and question
yes-or-no.

### DIFF
--- a/exwm-manage.el
+++ b/exwm-manage.el
@@ -533,9 +533,11 @@ manager is shutting down."
                                  exwm--connection)))
       (xcb:flush exwm--connection)
       (with-timeout (exwm-manage-ping-timeout
-                     (if (y-or-n-p (format "'%s' is not responding.  \
+                     (if (progn
+                           (exwm-input--grab-keyboard)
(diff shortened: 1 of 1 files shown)


exwm-manage.el
```diff
@@ -533,9 +533,11 @@ manager is shutting down."
                                  exwm--connection)))
       (xcb:flush exwm--connection)
       (with-timeout (exwm-manage-ping-timeout
-                     (if (y-or-n-p (format "'%s' is not responding.  \
+                     (if (let ((exwm-input-line-mode-passthrough t))
```
ch11ng: Remove it since it does nothing.

ch11ng:
Abandon in favor of 5b9f4b0851d27e8132a930dfbfec0b94099a4c9f.

PR #604, opened 2019-08-02
Use the wiki Configuration-Example for exwm-input-simulation-keys

a-schaefers:
This way, C-s and C-y works the way we expect in  applications such as web browsers, etc.


ch11ng:
I'm afraid some users may simply use the default simulation key configuration and this change can affect and confuse them.  Besides IMHO the CUA keys are less universal so I didn't make them default.  For instance <kbd>C-w</kbd> is handy for closing browser tabs and killing the previous word in terminal simulators.

medranocalvo:
I agree with @ch11ng. Nevertheless, it migth be helpful to new users to add them commented.

PR #664, opened 2019-09-29
Avoid entering line-mode after exiting ivy

Stebalien:
Extension of 02b1be7160051610a542764ec392e3a50f23fb5a.

Without this, the following command will swallow one key after it exits.

```bash
emacsclient -e '(ivy-read "prompt: " (list "foo" "bar"))'
```


ch11ng:
Ivy actually calls `exit-minibuffer` but we can't easily detect it.  Though widely used ivy is not a builtin so I'd rather make a blacklist for this.  Please checkout 7778766af824326d89cf4bce7541823218d43adc.

Stebalien:
That works for me. Thanks!

PR #701, opened 2020-01-08
Allow non-transparent systemtray background

timor:
When using light emacs themes, my icons used in the tray happen to be nearly invisible.

This change allows to configure a solid background color instead of the transparent background.

### DIFF
--- a/exwm-floating.el
+++ b/exwm-floating.el
@@ -758,34 +758,43 @@ Both DELTA-X and DELTA-Y default to 1.  This command should be bound locally."
                           nil nil))
     (xcb:flush exwm--connection)))
 
-(defun exwm-floating--init-border ()
-  "Initialize border colormap and pixel."
-  (exwm--log)
+(defun exwm-floating--init-colormap ()
+  "Initialize colormap used for border and systemtray background."
   ;; Use the default colormap.
+  (exwm--log)
   (unless exwm-floating--border-colormap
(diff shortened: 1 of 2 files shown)


ch11ng:
In X11 a pixel is a colormap-depended value.  In a colormap with 'TrueColor' visual class it's exactly the corresponding RGB value.  I'm not sure if 'TrueColor' is standard nowadays.  The 'tray' X window inherits the colormap of Emacs frame, which again seems to inherit that of the root X window, and the visual class of the root X window is configurable.  So I'm afraid your code won't work everywhere.

timor:
Ah I see.  Is there a way to get the correct value from X somehow?  It does not have to be an exact color, e.g. specifying black, white or gray would be sufficient for starters.

ch11ng:
There is an example: `exwm-floating-border-color`.  Perhaps it's better to create those X windows with 'TrueColor' visual class in the first place.

timor:
I changed it so that it follows what `exwm-floating-border-color` does.  Factored out a bit of code there to avoid duplicate code.



> Perhaps it's better to create those X windows with 'TrueColor' visual class in the first place.

Unfortunately, I don't know enough about X to have an educated opinion about that...

ch11ng:
It turns out colormaps other than 'TrueColor' are rarely used so I made the simplification to existing code and added background color to systemtray support BTW.  Please checkout the latest code.

Thanks for sharing the idea.  I'm closing this PR now.

PR #722, opened 2020-03-21
Fixed fail update when current buffer differs from window

Curiosidad-Racional:
Local variable `exwm--input-mode' from different buffer when current buffer don't contains #id window.

`with-current-buffer' changes buffer after `cl-case' with local `exwm--input-mode'


medranocalvo:
Thank you @Curiosidad-Racional, the change is right.

Let's wait for @ch11ng to apply it.

As an aside, please consider assigning copyright to the FSF for Emacs-related contributions if you've not done so yet, so that further/bigger contributions can be accepted in the future.

Thanks again,
Adrián.

PR #736, opened 2020-04-15
Abort recursive edit on workspace switch?

### LINKED ISSUE #733: Extra workspace switch when minibuffer active
Something like this was already noticed in the comments of #580.

If I have an active minibuffer (I tried ivy, helm, and just `eval-expression`), then switch workspaces, EXWM bounces me back to the original workspace. It also closes the minibuffer, so subsequent calls to `exwm-workspace-switch` work fine.

It's mildly annoying, but since I finally correlated it with the minibuffer being active, I thought I'd report it here.

To be clear, this is what I see.

In workspace 1: Run `eval-expression` and leave the minibuffer active. Then trigger `exwm-workspace-switch` with a key binding to switch to workspace 2.

EXWM then takes me from 1 -> 2, then immediately back 2 -> 1. The minibuffer is no longer active, and I can switch 1 -> 2 fine.

medranocalvo:
These changes intend to fix #733.

The first commit corrects the misbehaviour of `exwm-workspace-switch' by aborting the recursive edit as soon as possible and scheduling the workspace switch.

The second commit removes the whole behaviour.  The reason is that this situation is not different from non-EXWM Emacs when having an active minibuffer in one frame and switching to another frame.  Users should handle it likewise, for example setting `enable-recursive-minibuffers` or living with the "Command attempted to use minibuffer while in minibuffer" errors.

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -587,16 +587,6 @@ for internal use only."
     (when (or force (not (eq frame exwm-workspace--current)))
       (unless (window-live-p window)
         (setq window (frame-selected-window frame)))
-      ;; Close the (possible) active minibuffer.  Aborting the recursive edit
-      ;; level will abort the execution of this very command.  Schedule it to
-      ;; run immediately afterward.  This must all be done before the new
-      ;; workspace is selected in the new workspace, in order to avoid the focus
-      ;; to go back to the previous frame due to resetting the window
-      ;; configuration (e.g. `eval-expression').
-      (when (active-minibuffer-window)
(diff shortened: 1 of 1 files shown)


medranocalvo:
@ch11ng, please select the preferred approach.

medranocalvo:
@mindlike reported (#733):

> FYI I've been using commit cddb342 for a few days without any issues.

I myself have been using cddb342 without issues as well.

Unless there's a reason for keeping the behaviour, I'd say we should merge both commits.

medranocalvo:
Perfect, thanks.

Issue #705, opened 2020-01-30
exwm-input--update-focus cause key not sent to input-method-function


QiangF:
See the debug log here,
https://gist.github.com/QiangF/3da9f935354d31657a056ef31c6127cf

Every time exwm-input--update-focus is run, one key (118) is not processed by the input method,
the focused window is not changing. Is the input method disable during that instant?

QiangF:
I think the fix will be: if the focus did not change, don't run that function.

QiangF:
Fixed by https://github.com/ch11ng/exwm/pull/706

ch11ng:
That change may have some unintended side effects I'm afraid.  I wonder why there was a change in buffer-list.  I didn't find the definition of `rime-input-method` though.

QiangF:
rime-input-method is my simplified chinese input method:
https://gist.github.com/QiangF/d726722d4ceee0cd31fc59163ae2a5bd
I put exwm--log there to print the key received by input method. I have not seen any side effect by far.

There might be better fix if we know the reason for the issue.

QiangF:
The change in buffer-list is probably because I use a temporary buffer for the preedit. Also exwm-xim has a " *temp*" buffer.

ch11ng:
I've not tried it myself but it seems your input method does call functions like `get-buffer-create` which in turn run `buffer-list-update-hook`.

ch11ng:
What about let-bind `exwm-input--skip-buffer-list-update` to `t` so as to skip certain calls to `buffer-list-update-hook`? This should not be abused though.

QiangF:
No, "let-bind exwm-input--skip-buffer-list-update to t" doesn't work.
I tried set it both in exwm-xim:

             (let ((input-method-use-echo-area t)
                   (exwm-input--skip-buffer-list-update t))
                   (with-temp-buffer

and in my input method:

    (defmacro rime-with-preedit (&rest forms)
    (declare (indent 0) (debug t))
    `(let ((exwm-input--skip-buffer-list-update t))
        (with-current-buffer
            (get-buffer-create " *rime-preedit*")
        ,@forms)))

I have type 7 key 118, and the last one is lost, see the debug output:

https://gist.github.com/QiangF/721dec58a86405dc72fdde3dfade2c61

QiangF:
[...]

PR #706, opened 2020-01-31
fix key bypassing input-method-function during exwm-input--update-focus

QiangF:
fix https://github.com/ch11ng/exwm/issues/705

PR #737, opened 2020-04-15
Improve detection of focus changes

medranocalvo:
* exwm-input.el: (exwm-input--on-buffer-list-update): Keep track
	of last selected window and buffer, update focus only when those
	change.
	(exwm-input--update-focus-defer): Add commentary.
	(exwm-input--buffer-list-update-last-selected-window)
	(exwm-input--buffer-list-update-last-selected-buffer): Add
	variables.
	(exwm-input--skip-buffer-list-update): Remove variable.
	(exwm-input--on-buffer-list-update): Stop checking
	`exwm-input--skip-buffer-list-update'; it's no longer needed when
	keeping track selected window and buffer.

	* exwm-manage.el (exwm-manage--manage-window): Remove binding of
	`exwm-input--skip-buffer-list-update'.

This should limit substantially the number of `on-buffer-list-updates` we react to.  Especially, it should ignore all `with-temp-buffer`, as those don't select a different window (note that we use `(window-buffer (selected-window))` and not `(current-buffer)`).

A further improvement could be to check that the selected window belongs to an EXWM-managed frame (workspace, floating, minibuffer(?)).

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -138,7 +138,15 @@ defined in `exwm-mode-map' here."
   "Timer for deferring the update of input focus.")
 
 (defvar exwm-input--update-focus-window nil "The (Emacs) window to be focused.
+It also helps us discern whether a `buffer-list-update-hook' was caused by a
+different window having been selected.
+
 This value should always be overwritten.")
+
+(defvar exwm-input--update-focus-window-buffer nil
+  "Buffer displayed in `exwm-input--update-focus-window'.
+Helps us discern whether a `buffer-list-update-hook' was caused by the selected
+window switching to a different buffer.")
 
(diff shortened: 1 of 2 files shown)


medranocalvo:
Some notes: as mentioned in #706, I attributed part of the changes to @QiangF.  Mentioned there as well, this should fix #705.

Stebalien:
I haven't rigorously tested it but, while `exwm-layout--refresh-workspace` is still noticeable, this patch appears to help.

ch11ng:
@medranocalvo Do you think we should merge it anyway? It should help alleviate other problems at least.

bendlas:
Applied this to my fork https://github.com/bendlas/exwm

I think it fixes the focus problems, I've been having ..

Issue #739, opened 2020-04-22
super-save mode causes workspace switch


ghost:
I use [super-save mode](https://github.com/bbatsov/super-save) to automatically save buffers when they lose focus. If I enable this mode, the following unwanted sequence occurs:

1. Edit a buffer in workspace A
2. Switch to workspace B
3. The buffer in workspace A gets saved and EXWM automatically switches back to workspace A. 

FYI @bbatsov

medranocalvo:
Thanks for your report.

First please try applying the following patch:

```diff
diff --git c/exwm-workspace.el i/exwm-workspace.el
index f63a0a36e..cff17f3a1 100644
--- c/exwm-workspace.el
+++ i/exwm-workspace.el
@@ -587,6 +587,11 @@ for internal use only."
     (when (or force (not (eq frame exwm-workspace--current)))
       (unless (window-live-p window)
         (setq window (frame-selected-window frame)))
+    (when (and (not (eq frame old-frame))
+               (frame-live-p old-frame))
+      (with-selected-frame old-frame
+        (funcall exwm-workspace--original-handle-focus-out
+                 (list 'focus-out frame))))
       ;; Raise this frame.
       (xcb:+request exwm--connection
           (make-instance 'xcb:ConfigureWindow
@@ -680,11 +685,6 @@ for internal use only."
                              :dst-x (/ (frame-pixel-width frame) 2)
                              :dst-y (/ (frame-pixel-height frame) 2)))
           (xcb:flush exwm--connection))))
-    (when (and (not (eq frame old-frame))
-               (frame-live-p old-frame))
-      (with-selected-frame old-frame
-        (funcall exwm-workspace--original-handle-focus-out
-                 (list 'focus-out frame))))
     (funcall exwm-workspace--original-handle-focus-in (list 'focus-in frame))
     (run-hooks 'exwm-workspace-switch-hook)))
```

If that does not work (or breaks something else), please record the log (see https://github.com/ch11ng/exwm/wiki#how-to-report-a-bug, esp point 4. and 5.).  Before doing that, add, `(xcb-debug:backtrace)` after this line https://github.com/ch11ng/exwm/blob/master/exwm-input.el#L299.  Let's see if we can find the culprit.

Once that's done, you can try adding the workspace switching commands to `super-save-triggers`.

medranocalvo:
@mindlike, please try #744.

medranocalvo:
#744 is merged, and the reporter is MIA.  Closing.

PR #744, opened 2020-05-06
Prevent jumping back to previous workspace

medranocalvo:
* exwm-workspace.el (exwm-workspace-switch): Focus out old frame
	before switching to the new one to prevent jumping back to
	previous workspace.

Should help with #739.


ch11ng:
@medranocalvo You may push such small changes directly next time ;) Just don't forget GNU Savannah.

medranocalvo:
Thanks.  You already told me that a long time ago, but I kind-of prefer your review.  Perhaps next time.

Issue #845, opened 2021-07-29
Is Chris OK?


divansantana:
Hi Chris,

Just wondering if you are OK?

Your github page shows no activity since 06 2020.

Being from Hangzhou China and with the current world wide events, I'm
just wondering if you are alright?

TuLithu:
Does anyone have an email for this developer? If he does not respond, it may be necessary to fork this project in order to continue development. I use this software every day. It is my primary interface. I'm sure there are hundreds, if not thousands, that are also. It is too important to allow development to stop.

medranocalvo:
@divansantana, @TuLithu, I'll try contacting @ch11ng.

W.r.t. development: what's needed in order to continue development is... to continue development.  Up to now, @ch11ng has been the only one continuing development.  I have commit access to the repository, but have little time.

divansantana:
I did try contact Chris on Jul 15. No reply so far.

Hopefully he's just busy with some other life stuff.

zevlg:
> I did try contact Chris on Jul 15. No reply so far.
> 
> Hopefully he's just busy with some other life stuff.

Can we contact one of his colleague directly to find out the info about Chris? https://www.sea.com/aboutus/leadership/management/ChrisFeng

NinKenDo64:
I too have had my (self-entitled) mild frustrations begin turning into concern for the developer's well-being. I think in the interests in good taste it behoves us to ascertain his well-being before a fork.

divansantana:
> > I did try contact Chris on Jul 15. No reply so far.
> > Hopefully he's just busy with some other life stuff.
> 
> Can we contact one of his colleague directly to find out the info about Chris? https://www.sea.com/aboutus/leadership/management/ChrisFeng

He seems to be good and well then, I hope.

I guess if he wasn't the above link would be updated with a different CEO.

This article mentions 
https://www.sea.com/media/173 from Jan 25, 2021 mentions him as CEO still attending and speaking.  Nothing more recent.

Hopefully just busy with life.

medranocalvo:
[...]

PR #848, opened 2021-09-06
Use a hash table to cache result of ‘exwm-workspace--client-p’

mattbeshara:
Hi there,
I've been doing some CPU and memory profiling of my EXWM desktop, and the ‘exwm-workspace--client-p’ function conses a surprising amount of memory.  In one benchmark where I moved point from the beginning to the end of a screenful of text without any local minor modes enabled using only ‘right-char’, this function allocated nearly 4MB of memory.  Not much in the grand scheme of things, but any allocation that can be avoided should be in order to reduce GC times, and with the implementation in this PR, the memory allocated is reduced to <85k.

Thanks,
Matt

### DIFF
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -165,10 +165,19 @@ NIL if FRAME is not a workspace"
   "Return t if FRAME is a workspace."
   (memq frame exwm-workspace--list))
 
+(defvar exwm--client-p-hash-table
+  (make-hash-table :test 'eq :weakness 'key))
+
 (defsubst exwm-workspace--client-p (&optional frame)
   "Return non-nil if FRAME is an emacsclient frame."
-  (or (frame-parameter frame 'client)
-      (not (display-graphic-p frame))))
+  (let* ((frame (or frame (selected-frame)))
+         (cached-value (gethash frame exwm--client-p-hash-table 'absent)))
+    (if (eq cached-value 'absent)
+        (puthash frame
+                 (or (frame-parameter frame 'client)
+                     (not (display-graphic-p frame)))
+                 exwm--client-p-hash-table)
+        cached-value)))
 
 (defvar exwm-workspace--switch-map nil
   "Keymap used for interactively selecting workspace.")
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -1469,7 +1469,8 @@ the next workspace."
       ;; care of converting a workspace into a regular unmanaged frame.
       (let ((exwm-workspace--create-silently t))
         (make-frame)))
-    (exwm-workspace--remove-frame-as-workspace frame))))
+    (exwm-workspace--remove-frame-as-workspace frame)
+    (remhash frame exwm--client-p-hash-table))))
 
 (defun exwm-workspace--on-after-make-frame (frame)
   "Hook run upon `make-frame' that configures FRAME as a workspace."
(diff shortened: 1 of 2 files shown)


timor:
Interesting.  Do you have an idea about what exactly causes the consing?

mattbeshara:
I think it might be a call to `frame-parameters` inside the C implementation of `frame-parameter`.  I haven’t done the most thorough investigation, but `frame-parameters` does call `copy-alist`, which conses.  I don’t think it’s necessarily generating a lot of garbage each call, but `exwm-workspace--client-p` is called from both `exwm-input--on-echo-area-dirty` and `exwm-layout--on-echo-area-change`, and both of those functions appear to be getting called for every keystroke.

daviwil:
Yikes, really nice catch!  It'd be lovely to get this merged.

mattbeshara:
I think I read somewhere that the maintainer of this project seems to have abandoned it, so I'm not sure if this change actually has any chance of getting merged any time soon.  But in any case, I think this package is in GNU ELPA, so I'd like to confirm that I have completed the FSF copyright assignment paperwork and that that should not prevent merging this code if that becomes possible in future.

daviwil:
@medranocalvo Do you have the ability to merge PRs in this repo?

zackteo:
@mattbeshara Perhaps you were referring to this discussion - https://github.com/ch11ng/exwm/issues/845 :) but yes would be great if it could get merged

medranocalvo:
@mattbeshara, @daviwil: yes, I can merge PRs to ch11ng/exwm.  I truly want to get EXWM project rolling again (at least "life-support"), but struggle to find the time.  I've also been thinking how to handle the situation in a respectful way.

@mattbeshara: thank you very much for the patch.  It's great that you have signed the papers, especially w.r.t. to future contributions.  I will review it and get back to you as soon as possible.  Quick comments:

- Looks good
- Perhaps we could remove the frames from the cache on `kill-frame-hook` (e.g. `exwm-workspace--on-delete-frame`; not sure, just had a quick look).
- There are countless `frame-parameter` calls... I wonder which others are a hog, and whether we can do better.
- Should the `frame-parameter` slowness be reported as an Emacs bug?

daviwil:
@medranocalvo I would love to see this project become active again, I depend on it heavily!  I've started independently trying to diagnose and fix bugs in EXWM on my [live streams](https://www.youtube.com/watch?v=mTJ54yNFgZI&list=PLEoMzSkcN8oOS1y2uMspTXr1nd5JxUSzB) and I'd love to get some fixes merged back in once I finally get the FSF paperwork done.

mattbeshara:
@medranocalvo  I’ll push a new commit shortly which explicitly removes workspace frames which are being deleted from the cache.

It appears that `frame-parameters` cannot avoid calling `Fcopy_alist` because it does modify the list it copies, so I don’t think we can regard how much it conses to be buggy behaviour.  I haven't spent the time to understand the code 100%, but it doesn’t seem like it’s doing much it doesn’t need to do.

`frame-parameter` does already have special-case handlers for lookups of values of certain parameters in place to avoid consing, but I’m not sure if it would be viable to add more for EXWM.  Looking through the codebase, it seems that EXWM could probably be made more performant if it maintained its own data structures for keeping track of certain frame-specific bits of information instead of using frame parameters to do so.  One or more hash tables with frames as keys would seem to work just as well for that purpose as it does for the caching in this PR.

medranocalvo:
I have released 0.25 with latest @ch11ng changes.  Will release 0.26 with your changes ASAP after 0.25 is in GNU ELPA.

medranocalvo:
@mattbeshara: you are now an EXWM developer.  Congratulations!

Thank you for the improvements and also for your patience and persistence.  I'll get around your other pull request as time permits.

PR #850, opened 2021-10-02
Use a list rather than a frame property to track ‘urgent’ frames

mattbeshara:
Following the discussion in the comments in #848 here is a PR which reduces the number of times `frame-parameter` is called.  On my setup, `exwm-urgent` was the most frequently requested parameter out of those beginning with `exwm-` (`exwm-active` is the runner up).  I think this is because it was being used in mode line refreshes.

It turns out that this frame parameter corresponds to an attribute ('hint') of windows which in other WMs would blink their icon in the dock or something along those lines to try to attract attention.  I don't care about such functionality myself and I'm sure I've never even noticed the minor visual changes EXWM makes to indicate 'urgent' windows before.  Regardless, I've switched EXWM from storing this info in a frame property to using a list, and this should mean there are far fewer calls to `frame-property` being made.  As such, no memory should be consed by code simply checking the value of this property for a given frame.  Frames which have this property set are added to a list, and membership of that list is checked with `memq`.  Simple, but it should be effective.  I haven't been using these changes for long so I can't say if it makes any noticeable difference to performance, but in theory the fewer calls to `frame-property`, the better.

### DIFF
--- a/exwm-core.el
+++ b/exwm-core.el
@@ -348,9 +348,8 @@ One of `line-mode' or `char-mode'.")
   ;;
   (setq mode-name
         '(:eval (propertize "EXWM" 'face
-                            (when (cl-some (lambda (i)
-                                             (frame-parameter i 'exwm-urgency))
-                                           exwm-workspace--list)
+                            (when (cl-intersection exwm--urgent-frames
+                                                   exwm-workspace--list)
                               'font-lock-warning-face))))
   ;; Change major-mode is not allowed
   (add-hook 'change-major-mode-hook #'kill-buffer nil t)
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -286,8 +286,8 @@ NIL if FRAME is not a workspace"
                           (propertize
                            (apply exwm-workspace-index-map (list j))
                            'face
-                           (cond ((frame-parameter (elt exwm-workspace--list j)
-                                                   'exwm-urgency)
+                           (cond ((memq (elt exwm-workspace--list j)
+                                        exwm--urgent-frames)
                                   '(:foreground "orange"))
                                  ((aref not-empty j) '(:foreground "green"))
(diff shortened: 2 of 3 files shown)


mattbeshara:
@medranocalvo Would there be any interest in merging this?  I don't think I use any programs which ever set the `urgent` flag on windows, so I can't say that I've really done much in terms of testing it.  It seems fine, but the performance benefits it yields may be insignificant.

medranocalvo:
Hello @mattbeshara,

sorry for the long response times, I'm very busy at the moment.

Your first patch was very useful in showing the performance problem, and found `frame-parameter` as the costly cause.

Since then, I've modified my opinion on the matter: I now think that:

* Our use of `frame-parameter` in EXWM is correct: `frame-parameter` is the Emacs way of storing frame-local values.
* `frame-parameter` is unnecessarily slow for stored frame-local values.  I digged a little in Emacs' sources and, on a high level and as far as I remember, the cause is that the whole frame-parameter structure is build before searching our parameter; this includes querying X about various graphical properties.
* The right fix for this issue is to correct `frame-parameter` unnecessary slowness is Emacs proper.
* The slowness you found out was caused by the use of `frame-parameter` in a critical section of EXWM: a function attached to `buffer-list-update-hook`, which is called incessantly.  I have since then concluded that the call was unneeded. Please see https://github.com/ch11ng/exwm/pull/869.

Sorry for the terse message, I'm open for discussion and would gladly read your opinion.

Kind regards,
Adrián.

mattbeshara:
Hi Adrián,
No need to apologise!  You have clearly dug into the issue more extensively than I have, so if your view is that this PR is attempting to solve the problem in the wrong way, then I have no problem with that.  We can close this PR.  Is there a bug filed on Emacs itself for this issue with `frame-parameter`?

Issue #444, opened 2018-06-13
Support for gaps


agzam:
There are many reasons why I love EXWM. Despite all the quirks and bugs it is really a godsend, it is truly "dreams come true" for die-hard Emacsen. Thank you for making it possible. 

I personally don't care about this, but when I shared my excitement about EXWM on reddit - a few people responded, asking if it can support gaps. Some even said: "it's a deal-breaker for them, because it doesn't".

I wonder if this would require a ton of work? Probably it would have to be able to render multiple frames in the same screen etc, etc.

Anyhow I think it would be nice to have and it would be awesome to attract users who'd want this feature.

Stebalien:
IMO, this is more of an emacs issue than an EXWM issue. You can sort-of do this by customizing window dividers (make sure to enable window-divider-mode first): https://www.gnu.org/software/emacs/manual/html_node/elisp/Window-Dividers.html

However, that won't show you your X background. There may be a way to make this gap transparent but I haven't found it.

zevlg:
xwem had such functionality, so it might be possible in exwm as well.  xwem was possible to penetrate holes inside frames, so it could look like this http://lgarc.narod.ru/pics/xwem/xwem-2.1.png

the code for hole penetrator is at https://github.com/zevlg/xwem/blob/master/utils/xwem-holer.el

the idea is simple - just create special window above the frame, so you can see through

agzam:
> You can sort-of do this by customizing window dividers (make sure to enable window-divider-mode first)

My understanding was that to make this properly work - EXWM need to be able to render multiple **frames** side-by-side on the same screen, and be able to resize them to expose X11 background.

ch11ng:
I have no experience with such feature so it'd nice if someone could make a list of what to do or what things should be like.

With EXWM one usually do not have wallpaper set but with a compositor it makes sense (Emacs frames would be made semitransparent).  From @zevlg's example I think it's feasible to create fake transparent areas.

meedstrom:
A [good r/unixporn picture](https://www.reddit.com/r/unixporn/comments/g549vq/bspwm_studying_birds_migration_is_easier_with_a/) says a lot.

[...]

PR #858, opened 2021-12-12
Transparency and aesthetics support/ideas

djeis97:
This draft PR contains a handful of patches I made to workspace handling to
a) support emacs forks/patches which use 32-bit color depths and render with partial transparency
b) set WM_STATE on workspace frames, so compositors and the like recognize them properly
b) shrink workspace frames inward by the size of a window divider on each side

Together with the changes in https://github.com/djeis97/emacs/tree/transparency (themselves based on the wonderful alpha-background patches from https://github.com/TheVaffel/emacs) which make window dividers draw transparent, I have produced something akin to i3-gaps.

However, I'm sure there are better ways to accomplish some (if not all) of what I've done here and I'm not sure how much of this even makes sense to bring into exwm proper. If nothing else, finding a visual should probably be extended to gracefully handle the case of no 32-bit visual being available. Any thoughts/suggestions would be greatly appreciated.

emacs_transp

### DIFF
--- a/exwm-core.el
+++ b/exwm-core.el
@@ -193,6 +193,21 @@ least SECS seconds later."
               (lsh (lsh (pop rgb) -8) 8)
               (lsh (pop rgb) -8)))))
 
+(defun exwm--get-visual-depth-colormap (conn id)
+  "Get visual, depth and colormap from X window ID.
+Return a three element list with the respective results."
+  (let (ret-visual ret-depth ret-colormap)
+    (with-slots (visual colormap)
+        (xcb:+request-unchecked+reply conn
+            (make-instance 'xcb:GetWindowAttributes :window id))
+      (setq ret-visual visual)
+      (setq ret-colormap colormap))
+    (with-slots (depth)
+        (xcb:+request-unchecked+reply conn
+            (make-instance 'xcb:GetGeometry :drawable id))
+      (setq ret-depth depth))
+    (list ret-visual ret-depth ret-colormap)))
+
 ;; Internal variables
 (defvar-local exwm--id nil)               ;window ID
 (defvar-local exwm--configurations nil)   ;initial configurations.
--- a/exwm-workspace.el
+++ b/exwm-workspace.el
@@ -1326,7 +1326,8 @@ Please check `exwm-workspace--minibuffer-own-frame-p' first."
   (let ((outer-id (string-to-number (frame-parameter frame
                                                      'outer-window-id)))
         (window-id (string-to-number (frame-parameter frame 'window-id)))
-        (container (xcb:generate-id exwm--connection)))
+        (container (xcb:generate-id exwm--connection))
+        frame-colormap frame-visual frame-depth)
     ;; Save window IDs
     (set-frame-parameter frame 'exwm-outer-id outer-id)
     (set-frame-parameter frame 'exwm-id window-id)
@@ -1340,9 +1341,17 @@ Please check `exwm-workspace--minibuffer-own-frame-p' first."
       (dolist (param '(exwm-randr-monitor
                        exwm-geometry))
         (set-frame-parameter frame param (frame-parameter w param))))
+    ;; Support transparency on the container X window when the Emacs frame
+    ;; does.  Note that in addition to setting the visual, colormap and depth
+    ;; we must also reset the `:border-pixmap', as its default value is
+    ;; relative to the parent window, which might have a different depth.
+    (let* ((vdc (exwm--get-visual-depth-colormap exwm--connection outer-id)))
+      (setq frame-visual (car vdc))
+      (setq frame-depth (cadr vdc))
+      (setq frame-colormap (caddr vdc)))
     (xcb:+request exwm--connection
         (make-instance 'xcb:CreateWindow
-                       :depth 0
+                       :depth frame-depth
                        :wid container
                        :parent exwm--root
                        :x -1
@@ -1351,10 +1360,14 @@ Please check `exwm-workspace--minibuffer-own-frame-p' first."
                        :height 1
                        :border-width 0
                        :class xcb:WindowClass:InputOutput
-                       :visual 0
+                       :visual frame-visual
                        :value-mask (logior xcb:CW:BackPixmap
+                                           xcb:CW:BorderPixel
+                                           xcb:CW:Colormap
                                            xcb:CW:OverrideRedirect)
-                       :background-pixmap xcb:BackPixmap:ParentRelative
+                       :background-pixmap xcb:BackPixmap:None
+                       :border-pixel 0
+                       :colormap frame-colormap
                        :override-redirect 1))
     (xcb:+request exwm--connection
         (make-instance 'xcb:ConfigureWindow


daviwil:
Nice starting point, interested to see where this can possibly go!

medranocalvo:
Thank you very much for the patches.

As you see I’m not able to work much on EXWM. I’m sorry to have to ask you patience.

As I understand this is related to the transparency patches flying through emacs-devel these past weeks. They immediately reminded me of our request for gaps.

I’ll have a look as soon as I can.

Needless to say, I encourage everyone to test and/or comment on the changes.

Kind regards,
Adrián.

djeis97:
I've found an alternate way to produce the gap around the workspace frame- patch emacs' inner frame border to also be transparent. Since I'm relying on a transparency patch anyway, it seems better (for my workflow, at least) to extend that patch and rely on another existing emacs feature than try and shoehorn it into exwm. As such, I've dropped the last commit from this PR and now it's focused specifically on general transparency support in exwm.

jaor:
hi.  i've tried this code with the latest emacs master, and it seems to work well with the new alpha-background frame parameter.  however, i've encountered another problem when 32-bit is enabled: the system tray support in exwm stops working.  the effect is that when a new tray icon appears, a blank buffer gets created instead of it being displayed in the tray area.  blindly replacing :visual and :depth in exwm-systemtray.el with the new variables in this PR when they're 0 stops the buffers being created, but no icons are displayed :) -- sorry, i'm very ignorant when it comes to X11/xcb, so i don't really know what i'm doing! any chance that a real fix is not too complex? thanks!

djeis97:
What do you mean by “when 32 bit is enabled”? In an emacs build for a 32-bit architecture?

I may have some time to dig in to this this weekend- I haven’t tested the feature on the emacs master branch and I’ll probably need to put together a dedicated exwm config for it, but it’s probably high time I revisited this anyway.

jaor:
On Thu, Feb 03 2022, Elijah Malaby wrote:

> What do you mean by “when 32 bit is enabled”? In an emacs build for a
> 32-bit architecture?

No, in an emacs built from a recent checkout of its git master branch,
(where the new alpha-background functionality has been implemented),
compiled with either --with-x-toolkit=no or --with-x-toolkit=gtk.  For
those builds, the visual class is set to TrueColor-32 (so i was
referring to the number of visual planes, or the :depth parameter), and
that breaks systemtray.  Moreover, since depth 32 is needed for
alpha-background to work, the latter feature doesn't work either in
exwm.

If one sets in .Xdefaults

      Emacs.visualClass: TrueColor-24

then systemtray works again, but alpha-background doesn't (the
transparency applied to the background shows as if below on always had a
black background, ignoring the wallpaper image).

> I may have some time to dig in to this this weekend- I haven’t tested
> the feature on the emacs master branch and I’ll probably need to put
> together a dedicated exwm config for it, but it’s probably high time I
> revisited this anyway.

That'd be great.  With this PR as it is, alpha-background, the new
feature, works well for me (so using depth 32 is working for workspace
frames), but the system tray is still broken.  I think that is because
the system tray windows are still created with :depth 0, but, as i
mentioned, my attempts to naively fix that in exwm-systemtray.el by
mimicking what is done for workspaces haven't been succesful.

Thanks a lot for looking into this, and let me know if i can be of
further assistance!

medranocalvo:
@djeis97, I finally tested your changes with Emacs' master (b0fd3ec2c0131e79b1c62e1f1beb6f29cf076120).  Works great!

The second commit https://github.com/ch11ng/exwm/pull/858/commits/426c9115c4e22d1908a73049d83371b45f7c9a09 is a bug fix, I'll cherry-pick it.

With regards to the https://github.com/ch11ng/exwm/pull/858/commits/78d02a487e8f99b0c46b34a357fdbe4471569831, I worry about the situation where no 32-bit color depth visual can be found.  Would it be possible to reuse whatever visual and colormap the Emacs' frame is using?

[...]

After a quick test:

```elisp
(setq reply (xcb:+request-unchecked+reply exwm--connection
                (make-instance 'xcb:GetWindowAttributes
                               :window (string-to-number (frame-parameter nil 'outer-window-id)))))
(slot-value reply 'visual)
(slot-value reply 'colormap)
```

We can query Emacs' frame's visual and colormap.  I did not manage to query the depth, though I think some method of xcb:renderutil would help.

As far as I can see this is your first contribution to EXWM (thanks!).  Do you know about and have you completed the copyright assignment?

---

Gaps has been requested before #444.  If I understand correctly, Emacs must be patched so that the window dividers can be made transparent.  It would be great if you could submit your patches to that effect to Emacs.

As for the outside gap, it reminded me of the `internal-border-width` frame parameter.  I tested it and no, it does not become transparent when setting `alpha-background`.  Perhaps you could bring it up in emacs-devel, while the transparency theme is hot. Otherwise `exwm-randr` could be misused to do it (I have some WIP patches that would make it easier).

---

For the record, these are the steps I followed:

First compile Emacs master.

```sh
exit_hook() {
        kill -9 $XEPHYR_PID;
        kill -9 $EMACS_PID;
        kill -9 $COMPTON_PID;
}
trap exit_hook EXIT;

XEPHYR=Xephyr
XEPHYR_DISPLAY=:123

COMPTON=compton

EMACS=../emacs-master/src/emacs
EXWM_DIR=.
XELB_DIR=../xelb

$XEPHYR -resizeable $XEPHYR_DISPLAY & XEPHYR_PID=$?
export DISPLAY=$XEPHYR_DISPLAY

sleep 2s

$COMPTON & COMPTON_PID=$?
$EMACS --debug-init -L $XELB_DIR -L $EXWM_DIR & EMACS_PID=$?

# xsetroot not compatible with compton
hsetroot -solid '#0000ff'

echo 'Press Enter to exit...'
read _
```

Once Emacs starts up:

```elisp
(exwm-init)
(set-frame-parameter nil 'alpha-background 50)
```

medranocalvo:
@jaor, is the issue you describe the same as #863?

jaor:
On Wed, Feb 09 2022, medranocalvo wrote:

> @jaor, is the issue you describe the same as #863?

yes, i think so.

djeis97:
Apologies for the silence on this, I'll update the PR today. As for the internal border, the patches on my transparency branch (https://github.com/djeis97/emacs/tree/transparency) also make that transparent and that does seem to do the trick. I still need to test everything against emacs master, I've been a bit side-tracked for the last while. For the moment, for stability reasons, I've mostly been running my patches against the 28 release branch.

medranocalvo:
Hello @djeis97,  I've finally got some time to work on this.

I reworked your changes so that the container X window reuses the Emacs' frame's visual, depth and colormap.

I could not find your GNU copyright assignment.  Would you be willing to assign the copyright to the FSF?  I kindly ask you to consider it, as I will need it to incorporate your contributions to EXWM.

I already committed the WM_STATE patch with `Copyright-paperwork-exempt`.  For these changes, I am inclined to consider my patch a reimplementation of your changes (taking advantage of your key insight of setting the `depth`, `visual`, `colormap`, `background-pixmap` and `border-pixel` to overcome the BadMatch errors).  If you agree, I could commit my patch along with appropriate attribution and gratitude to you.  Given our apparently shared lack of time, this might be the fastest way to release this improvement.

How would you like to proceed?

Thank you again.

djeis97:
> I could not find your GNU copyright assignment. Would you be willing to assign the copyright to the FSF? I kindly ask you to consider it, as I will need it to incorporate your contributions to EXWM.

I will have to consider that going forward, but it is indeed not something I have gotten around to as of yet.

> For these changes, I am inclined to consider my patch a reimplementation of your changes (taking advantage of your key insight of setting the depth, visual, colormap, background-pixmap and border-pixel to overcome the BadMatch errors). If you agree, I could commit my patch along with appropriate attribution and gratitude to you. Given our apparently shared lack of time, this might be the fastest way to release this improvement.

This makes perfect sense to me, go for it. I just want exwm to support the feature :)

medranocalvo:
I'll adjust the commit message and merge as time permits.

Thank you.

P.S.: this package might help with the outer gaps: https://github.com/lucasgruss/exwm-outer-gaps.  I have not tried it.

medranocalvo:
Merged! :-)

I'll wait for #867 before releasing 0.27.

Thank you again, hope to see again you around here.

medranocalvo:
@djeis97: 0.27 has been released, featuring your changes.  Thank you very much!

PR #867, opened 2022-03-20
Use default visual, depth and colormap in systray embedder window

medranocalvo:
We were using the Emacs' frame's depth, but not the visual nor colormap.
This failed with Emacs 29 and its support for 32-bit depths.  We now use
the default screen's values in order to increase compatibility.

* exwm-core.el (exwm--get-visual-depth-colormap): New function.
* exwm-systemtray.el (exwm-systemtray--init): Use root window's
visual, depth and colormap. Reset all attributes that refer
(perhaps due to defaults) to the parent window, as it might have a
different visual, depth or colormap.
(exwm-systemtray--init): Set _NET_SYSTEM_TRAY_VISUAL.
(exwm-systemtray-background-color): Don't use ParentRelative
pixmap, as the visual, depth or colormap might be incompatible.

### DIFF
--- a/exwm-systemtray.el
+++ b/exwm-systemtray.el
@@ -1,7 +1,7 @@
 ;;; exwm-systemtray.el --- System Tray Module for  -*- lexical-binding: t -*-
 ;;;                        EXWM
 
-;; Copyright (C) 2016-2021 Free Software Foundation, Inc.
+;; Copyright (C) 2016-2022 Free Software Foundation, Inc.
 
 ;; Author: Chris Feng <chris.w.feng@gmail.com>
 
@@ -30,6 +30,7 @@
 
 ;;; Code:
 
+(require 'xcb-ewmh)
 (require 'xcb-icccm)
 (require 'xcb-xembed)
 (require 'xcb-systemtray)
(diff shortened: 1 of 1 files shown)


medranocalvo:
I plan on reverting [[DRAFT] Use 32-bit depth & visual](https://github.com/ch11ng/exwm/pull/867/commits/057d1bc35e8c5db4e88d5d05a9bc6b621c747b52) and disable transparent systray when using 32-bit dephs.  The reason is that it requires a more involved implementation and I'm can't do it at the moment.  See https://specifications.freedesktop.org/systemtray-spec/systemtray-spec-latest.html#visuals for details:

>_NET_SYSTEM_TRAY_VISUAL
>_NET_SYSTEM_TRAY_VISUAL visual_id VISUALID/32
>The property should be set by the tray manager to indicate the preferred visual for icon windows. To avoid ambiguity about the colormap to use this visual must either be the default visual for the screen or it must be a TrueColor visual. If this property is set to a visual with an alpha channel, the tray manager must use the Composite extension to composite the icon against the background using PictOpOver.

medranocalvo:
These new commits go back to using the default visual, depth & colormap (that of the root window) for the system tray window.  The "transparency" trick is only supported when Emacs uses a compatible visual (i.e. 24-bit).

* `exwm-systemtray-background-color' accepts the symbol `transparent` to make it transparent.  We also silently understand `nil` to mean transparent, as before.
* `exwm-systemtray-background-color'  accepts the symbol `workspace-background` to mean: use the background color of current workspace.  This seems alright, because the systemtray is reparented to current workspace on workspace switch.

I have only tested this lightly, as I can't easily use systemtray in my setup.  Please test.

---

Now that we have `workspace-background`, I'm not sure `transparent` is worth keeping at all.  As far as I can tell, only the background pixmap of the Emacs frame was used: there was no actual transparency.  The background pixmap seems to be... the background color.

Stebalien:
Works modulo https://github.com/ch11ng/exwm/pull/894.

medranocalvo:
This change is included in EXWM 0.27.

PR #868, opened 2022-03-20
Create `exwm-minor-mode`

medranocalvo:
With this change I intend to convert EXWM into a global minor mode.

Please check out and test.

Next steps would be e.g. creating a menu for the minor mode where workspaces can be managed (at the moment the menu is only available when an EXWM-mode buffer is current).  Please discuss further ideas.

### DIFF
--- a/exwm-config.el
+++ b/exwm-config.el
@@ -73,7 +73,7 @@
             ([?\C-d] . [delete])
             ([?\C-k] . [S-end delete]))))
   ;; Enable EXWM
-  (exwm-enable)
+  (exwm-minor-mode +1)
   ;; Configure Ido
   (exwm-config-ido)
   ;; Other configurations
--- a/exwm.el
+++ b/exwm.el
@@ -71,6 +71,7 @@
 (require 'exwm-floating)
 (require 'exwm-manage)
 (require 'exwm-input)
+(require 'seq)
 
 (defgroup exwm nil
   "Emacs X Window Manager."
@@ -109,6 +110,10 @@
   "Name of the subordinate Emacs server.")
 
 (defvar exwm--server-process nil "Process of the subordinate Emacs server.")
+
+(define-error 'exwm-frame-not-x "Frame not running under X environment" 'error)
+(define-error 'exwm-already-running "EXWM already running" 'error)
+(define-error 'exwm-other-wm "Other window manager running" 'error)
 
 (defun exwm-reset ()
   "Reset the state of the selected window (non-fullscreen, line-mode, etc)."
@@ -603,7 +608,7 @@
           selection (slot-value obj 'selection))
     (when (and (eq owner exwm--wmsn-window)
                (eq selection xcb:Atom:WM_S0))
-      (exwm-exit))))
+      (exwm--exit))))
 
 (defun exwm--init-icccm-ewmh ()
   "Initialize ICCCM/EWMH support."
@@ -761,7 +766,7 @@ manager.  If t, replace it, if nil, abort and ask the user if `ask'."
       (when (eq replace 'ask)
         (setq replace (yes-or-no-p "Replace existing window manager? ")))
       (when (not replace)
-        (user-error "Other window manager detected")))
+        (signal 'exwm-other-wm "Other window manager detected")))
     (let ((new-owner (xcb:generate-id exwm--connection)))
       (xcb:+request exwm--connection
           (make-instance 'xcb:CreateWindow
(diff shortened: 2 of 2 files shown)


medranocalvo:
A minor mode is some feature that can be enabled and disabled.  When enabled, it adds to or alters Emacs' capabilities, binds commands to some keys, shows some menus, etc.  This fits EXWM functionality.  Indeed, EXWM does all of this already, some of it in some peculiar ways.  I'd like to make EXWM less peculiar (where possible), and this is a small step in that direction.

There are no functionality changes, though some of the entrypoints (e.g. `exwm-enable`) become obsolete.

ArneBab:
Can exwm as minor mode restore the previous window manager (if there was one)?

Stebalien:
> Can exwm as minor mode restore the previous window manager (if there was one)?

That's generally not possible to do reliably. Many window managers are components of larger systems.

ArneBab:
It seems to be possible with xprop: https://askubuntu.com/a/1153698/329876

`id=$(xprop -root -notype | awk '$1=="_NET_SUPPORTING_WM_CHECK:"{print $5}'); xprop -id "${id}" -notype -f _NET_WM_NAME 8t | grep "_NET_WM_NAME = " | cut --delimiter=' ' --fields=3 | cut --delimiter='"' --fields=2`

I’m asking because if I were to disable the minor-mode, I would not have a wm and that could cause Emacs to lose keyboard focus, leaving me without an ability to start another wm.

Would it maybe be possible to do this with a fallback? If there was a wm before (exwm already detects that) then use the detected wm and if that fails (or the previous one cannot be detected) use a fallback wm?

Stebalien:
That returns the window manager's name but that's not enough to _restore_ it. EXWM would need to know how to start the window manager and _most_ window managers expect to run within a specific environment (e.g., managed by a session manager).

ArneBab:
Aren’t you usually already in that environment if exwm replaced the window manager? (just starting the window manager by hand usually worked for me, if I could find out which one was used by the session manager)

Stebalien:
Not necessarily.

- There's no reliable way to get the exact command that started the
window manager. It may be possible to get the PID of the window manager
then get the command from /proc, but that's kind of hacky and isn't
guaranteed to work.

- There's no reason a window manager can't have multiple processes.
I.e., the window manager process (from the perspective of Xorg) may not
actually be the "root" window manager process.

In practice, it may be possible to "restart" most "simple" window
managers. But I wouldn't expect it to be reliable enough to actually
build it in as a feature. I.e., if you replace a running Gnome window
manager with EXWM, there's no way EXWM would be able to put everything
back together.

medranocalvo:
Indeed, we could read the previous WM name before replacing it.  I see the following challenges:

 1. Finding out the command needed to start the previous WM again.  We'd need to compile a dictionary matching the WM name to the command needed for launching it.  This research seems laborious to me, but I think it would be worthwhile if we succeed in restarting some frequently used WMs.  I implemented the replace functionality to ease trying out EXWM without the need to modify .xinitrc files; being able to restore the environment after trying would be a step further in that direction.  As @Stebalien points out, this might be impracticable.
 2. Invoking that command in such a way that it's not killed when exiting Emacs.
 3. ???

@ArneBab, would you be willing to work on this?  I have very limited time, but would try to support you as possible.  Please note that EXWM is part of GNU Emacs, so copyright assignment to the FSF would be needed.

At last, I'd like to point out that restarting the previous WM is orthogonal to this particular change: although a minor mode leads itself to being activated and deactivated, `exwm-exit` existed previously.

Thank you.

ArneBab:
@medranocalvo I have no problem with copyright assignment to GNU — can you send me the papers? (do you have my email address?)

This would make temporary exwm sessions more practical for me, so I would try to work on this (I likely won’t be fast, though, already too many projects on my plate).

My experience is that just starting another wm worked well enough for me. The hardest challenge was to remember or find the name of the window manager (or just start another one, they may not be fully interchangeable, but just having any window manager already makes most desktop environments work again).

medranocalvo:
> @medranocalvo I have no problem with copyright assignment to GNU — can you send me the papers? (do you have my email address?)

Please, have a look at https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html.

> This would make temporary exwm sessions more practical for me, so I would try to work on this (I likely won’t be fast, though, already too many projects on my plate).

Tell me about time!  I suggest you to open a PR with whatever first questions you might have.  Have a look at `exwm--wmsn-acquire`; perhaps a new option `restore` to `exwm-replace`...

> My experience is that just starting another wm worked well enough for me. The hardest challenge was to remember or find the name of the window manager (or just start another one, they may not be fully interchangeable, but just having any window manager already makes most desktop environments work again).

I tested various WMs while developing replace support (https://github.com/ch11ng/exwm/pull/380).  For me their `--replace` functionality (which not all of them support) was important, but it's not needed for our purposes: we can shut down EXWM before restarting the old WM.

medranocalvo:
I just remembered that I had issues using EXWM replace functionality under GNOME last time I tried.  When replacing the WM, I got a gigantic dialog whose only button would kill the session.  What I remember from my (non) understanding at the time is that unless the WM registers with the login manager (?) it will consider the session idle (?) and terminate it.  @Stebalien, is something like that what you had in mind?

See https://man.archlinux.org/man/systemd-logind.service.8.en.

Stebalien:
That specific case is probably not an issue in this case if EXWM has
already successfully replaced an existing window manager.

I was more concerned with the fact that restoring a window manager isn't
always as simple as starting a process. E.g.:

- In my case, I run Emacs within a systemd user service so any window
  managers I automatically spawn would die with emacs.
- Window managers may expect other components to be started alongside
  them and said components may have shutdown when the window manager was
  killed.

medranocalvo:
@Stebalien: thank you for your explanation.  You are right: we won't be able to recover to custom setups.  I'm also not sure whether we will be able to recover to complex standard setups, say GNOME4 with the login manager.  The objective is, as far as I'm concerned, making it seamless to try out EXWM (@ArneBab has a different but compatible objective, if I understood correctly).

@ArneBab: the following let's us recover XFCE4 after replacing it with EXWM.

```elisp
(call-process "xfwm4" nil 0 nil "--replace")
```
Seems to continue working after killing Emacs after replacing EXWM with XFWM4.

ArneBab:
@medranocalvo This would already make it much more painless to try exwm. Currently when I quit Emacs, I have to somehow start another window manager, which often means closing window after window until I finally get keyboard focus on a shell. Getting rid of that would go a long way to make a quick test more enjoyable.

medranocalvo:
I want to note that `exwm-mode` is ready, as far as I remember.  I'll rebase and post a new PR in @emacs-exwm.

I find the restarting the replaced WM is important feature for trying out EXWM.  I'll open a ticket on @emacs-exwm and will try to work on it (no promises).

minad:
Thanks, I like this change. I wonder if we should also change the scheme for the extensions, introducing `exwm-xsettings-mode` instead of `exwm-xsettings-enable` etc. What do you think?

PR #878, opened 2022-07-26
Hide tab-bar on floating windows

Stebalien:
Floating windows are dedicated to a specific buffer anyways.

### DIFF
--- a/exwm-floating.el
+++ b/exwm-floating.el
@@ -161,6 +161,8 @@ context of the corresponding buffer."
                           (get-buffer "*scratch*")))
                   (make-frame
                    `((minibuffer . ,(minibuffer-window exwm--frame))
+                     (tab-bar-lines . 0)
+                     (tab-bar-lines-keep-state . t)
                      (left . ,(* window-min-width -10000))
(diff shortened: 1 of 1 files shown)


medranocalvo:
I'll merge this ASAP.  Thank you very much.

I have not yet experimented with the tab-bar, but superficially seems like a good fit for EXWM workspaces.  One can even have the same X-Window on (different) tabs!  I wonder whether we should use them for workspaces and leave frames for floating windows and multi-monitor...

chasecaleb:
> One can even have the same X-Window on (different) tabs! I wonder whether we should use them for workspaces

If that's true, then I say absolutely yes! That's a huge improvement. This is the one dealbreaker with EXWM workspaces that keeps me from using them.

Stebalien:
> I wonder whether we should use them for workspaces and leave frames for floating windows and multi-monitor...

This is what I currently do and it works great.

medranocalvo:
Merged into master.  Will be released once we are happy with #867.

Thank you very much!

medranocalvo:
0.27 has been released, including this fix.  Thanks!

PR #879, opened 2022-07-26
Add a module for setting the X background

Stebalien:
When enabled, EXWM will automatically set the background color either the user-specified color, or theme's default background color.

This will even work with compositors such as picom (which is, unfortunately, where most of the complexity comes from).

### DIFF
--- a/exwm-background.el
+++ b/exwm-background.el
@@ -1,0 +1,201 @@
+;;; exwm-background.el --- X Background Module for EXWM  -*- lexical-binding: t -*-
+
+;; Copyright (C) 2022 Free Software Foundation, Inc.
+
+;; Author: Steven Allen <steven@stebalien.com>
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; This module adds X background color setting support to EXWM.
+
+;; To use this module, load and enable it as follows:
+;;   (require 'exwm-background)
+;;   (exwm-background-enable)
+;;
+;; By default, this will apply the theme's background color. However, that
+;; color can be customized via the `exwm-background-color' setting.
+
+;;; Code:
+
+(require 'exwm-core)
+
+(defcustom exwm-background-color nil
+  "Background color for Xorg."
+  :type '(choice
+          (color :tag "Background Color")
+          (const :tag "Default" nil))
+  :group 'exwm
+  :initialize #'custom-initialize-default
+  :set (lambda (symbol value)
+         (set-default-toplevel-value symbol value)
+         (exwm-background--update)))
+
+(defconst exwm-background--properties '("_XROOTPMAP_ID" "_XSETROOT_ID" "ESETROOT_PMAP_ID")
+  "The background properties to set.
+We can't need to set these so that compositing window managers can correctly display the background
+color.")
+
+(defvar exwm-background--connection nil
+  "The X connection used for setting the background.
+We use a separate connection as other background-setting tools may kill this connection when they
+replace it.")
+
+(defvar exwm-background--pixmap nil
+  "Cached background pixmap.")
+
+(defvar exwm-background--atoms nil
+  "Cached background atoms.")
+
+(defun exwm-background--update (&rest _)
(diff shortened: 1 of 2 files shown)


exwm-background.el
```diff
@@ -0,0 +1,201 @@
+;;; exwm-background.el --- X Background Module for EXWM  -*- lexical-binding: t -*-
+
+;; Copyright (C) 2022 Free Software Foundation, Inc.
+
+;; Author: Steven Allen <steven@stebalien.com>
+
+;; This file is part of GNU Emacs.
```
Stebalien, line 7: (I've signed the copyright assignment paperwork)

exwm-background.el
```diff
@@ -0,0 +1,201 @@
...
+                         :data
+                         (funcall (if xcb:lsb
+                                      #'xcb:-pack-u4-lsb
+                                    #'xcb:-pack-u4)
+                                  exwm-background--pixmap))))
+
+    ;; Kill the old background clients.
+    (dolist (pixmap old)
+      (xcb:+request exwm-background--connection
+          (make-instance 'xcb:KillClient :resource pixmap))))
```
Stebalien, line 139: Every wallpaper tool I've seen does this. This is why I open a new connection, and check to make sure it's still valid before changing the background.

exwm-background.el
```diff
@@ -0,0 +1,201 @@
...
+                       :pid exwm-background--pixmap
+                       :drawable exwm--root
+                       :width 1 :height 1))))
+
+(defun exwm-background--init ()
+  "Initialize background module."
+  (exwm--log)
+
+  (add-hook 'enable-theme-functions 'exwm-background--update)
+  (add-hook 'disable-theme-functions 'exwm-background--update)
```
Stebalien, line 177: This is a no-op before Emacs 29. Those versions won't react to theme changes, but that's not a _huge_ loss, IMO.

exwm-background.el
```diff
@@ -0,0 +1,201 @@
...
+(defun exwm-background--update (&rest _)
+  "Update the EXWM background."
+
+  ;; Always reconnect as any tool that sets the background may have disconnected us (to force X to
+  ;; free resources).
+  (exwm-background--connect)
+
+  (let ((gc (xcb:generate-id exwm-background--connection))
+        (color (exwm--color->pixel (or exwm-background-color
+                                       (face-background 'default)))))
```
Stebalien, line 73: I'm not sure if this is the right color. Ideally we'd pull out the frame background, but will that likely matter in practice?

Stebalien:
For a bit of motivation:

- This helps reduce flashes when restarting emacs, expanding to a new monitor, etc.
- This ensures that tray icons have the correct background, even in the presence of a compositor such as picom.

exwm-background.el
medranocalvo: Looking into it, `set-background-color` sets the frame background by modifying its frame parameters.  Internally it ends up changing the default face's background color, but only for that frame!  This module supports a single global background color, not one per frame.  Although changing the root background color when switching workspaces seems to be a useful feature, I don't think it's required.

exwm-background.el
medranocalvo: Yes, that seems correct.  Thank you for the explanation and especially the code comment above.

exwm-background.el
medranocalvo: No, it's the way it is.  We should update the Wiki with a recipe for advising `load-theme` to update background.  That should be enough.

exwm-background.el
medranocalvo: That's wonderful news :-).  Thank you.

medranocalvo:
This looks great, thank you very much.  I'll test it on my setup (have not done yet) and prepare the Wiki instructions.

medranocalvo:
>> I decided I'll disable systray transparency when using 32-bit depth emacs, reason in https://github.com/ch11ng/exwm/pull/867. This ticket will be turned into a feature request. Opinions?

> That seems reasonable for most EXWM users. It also reduces the need for https://github.com/ch11ng/exwm/pull/879.

I'm not sure I understand why is the need reduced.  Is it related to using 32-bit Emacs with a compositor?

---

I have not yet had the time to test it, I'm sorry.

Stebalien:
> I'm not sure I understand why is the need reduced. Is it related to using 32-bit Emacs with a compositor?

My experience is that, with 32-bit backgrounds, tray icons will take on the _wallpaper_ as the background color. This patch helped there by setting a wallpaper background.

> I have not yet had the time to test it, I'm sorry.

Np. This patch is, honestly, very low priority. It's easy enough to replicate with `hsetroot` and friends.

medranocalvo:
The following should be written in the Wiki (mostly copied from the file header):

    ## Root Background
    
    EXWM provides a way to set the root background color to match Emacs' theme or a fixed color.
    It's disabled by default and can be enabled with the following lines:

    ```elisp
    (require 'exwm-background)
    (exwm-background-enable)
    ```

    By default, this will apply the theme's background color.  However, that color can be customized via the `exwm-background-color' setting.

Additionally, we should add it to the `_Sidebar`:
```
  + [(Optional) Root Background](../wiki#root-background)
```

I'm not sure whether *X Background* or *Root Background* is more understandable.

medranocalvo:
@Stebalien, I squashed both commits into one and merged to EXWM's master branch.  Thank you very much!

Issue #889, opened 2022-10-10
A recent commit in emacs-master breaks point focus


Sbozzolo:
I updated my machine to Emacs master and found that I could no longer move the point across monitors (I'd have to use the mouse to click on the target monitor; the point icon would not move, even if I could type in the target buffer). The offending commit is between `07e6bbb9bc` and `07e6bbb9bc`.

I noticed that Po Lu did some work in this area (e.g., `5e7e85af02`, "Stop passing CurrentTime to SetInputFocus"). I quickly tried to bisect, but didn't have time to find the problematic commit. If this is up to upstream to fix, please let me know and I will report it.

mgi:
Hi,

I'd be happy to hear if #890 fixes your issue.

Best regards,

Sbozzolo:
Yes, it works.

mgi:
Gabriele Bozzola ***@***.***> writes:

> Yes, it works.

Thanks, we'll see if it makes it into EXWM then.
-- 
Manuel Giraud

Sbozzolo:
Actually, the fix is not perfect. Sometimes (I still have to find out when exactly), the focus moves, but the icon for the point does not show up. I can interact as it was there, but I do not see it.

Will report later if I find the specific conditions that trigger this behavior.

mgi:
Gabriele Bozzola ***@***.***> writes:

> Actually, the fix is not perfect. Sometimes (I still have to find out
> when exactly), the focus moves, but the icon for the point does not
> show up. I can interact as it was there, but I do not see it.
>
> Will report later if I find the specific conditions that trigger this
> behavior.

Yes, that would be very helpful.  I cannot see this but my usage of EXWM
is very simple (just one monitor and almost no usage of workspaces).
-- 
Manuel Giraud

yusi1:
I've been trying out your PR too and for me, it is fixed in `find-file` when in X windows (focus is moved to the minibuffer), but not something like `consult-buffer` which I use to switch buffers using a menu in the minibuffer, where it stays hollow when an X window is selected.

mgi:
Yusef Aslam ***@***.***> writes:

> I've been trying out your PR and for me, it is fixed in `find-file`
> when in X windows (focus is moved to the minibuffer), but not
> something like `consult-buffer` which I use to switch buffers using a
> menu in the minibuffer, where it stays hollow when an X window is
> selected.

Ok, I'm not a consult user myself but I've tested it and can reproduce
[...]

PR #890, opened 2022-10-12
Fix frame focus for Emacs 29

mgi:
See https://debbugs.gnu.org/cgi/bugreport.cgi?bug=58438 for more details

### DIFF
--- a/exwm.el
+++ b/exwm.el
@@ -420,8 +420,8 @@
     (setq type (slot-value obj 'type)
           id (slot-value obj 'window)
           data (slot-value (slot-value obj 'data) 'data32))
-    (exwm--log "atom=%s(%s)" (x-get-atom-name type exwm-workspace--current)
-               type)
+    (exwm--log "atom=%s(%s) id=#x%x data=%s" (x-get-atom-name type exwm-workspace--current)
+               type (or id 0) data)
     (cond
      ;; _NET_NUMBER_OF_DESKTOPS.
      ((= type xcb:Atom:_NET_NUMBER_OF_DESKTOPS)
@@ -443,7 +443,8 @@
      ((= type xcb:Atom:_NET_ACTIVE_WINDOW)
       (let ((buffer (exwm--id->buffer id))
             iconic window)
-        (when (buffer-live-p buffer)
+        (if (buffer-live-p buffer)
+          ;; Either an `exwm-mode' buffer (an X window) or a floating frame.
           (with-current-buffer buffer
             (when (eq exwm--frame exwm-workspace--current)
               (if exwm--floating-frame
@@ -457,7 +458,11 @@
                 (setq window (get-buffer-window nil t))
                 (when (or iconic
                           (not (eq window (selected-window))))
-                  (select-window window))))))))
+                  (select-window window)))))
+          ;; A workspace.
+          (dolist (f exwm-workspace--list)
+            (when (eq id (frame-parameter f 'exwm-outer-id))
+              (x-focus-frame f t))))))
      ;; _NET_CLOSE_WINDOW.
      ((= type xcb:Atom:_NET_CLOSE_WINDOW)
       (let ((buffer (exwm--id->buffer id)))


medranocalvo:
Hello @mgi,

I've read the discussion on #889.  In my opinion this fix is correct and necessary, although there appear to be further issues in the area.

I added a new commit that reverts back to your first version modifying it a bit.  Please have a look and modify as you see fit.

If you are OK with the changes (incl. commit message Thanks-To, which might be more personal), I propose that you rebase & commit under your name, if that's OK for you.

Contributions to EXWM usually need copyright assignment to the FSF.  Please let me know (privately if you prefer) whether you have already assigned it for changes to Emacs.  If not, please consider doing it, as I must require it for further contributions.  In other words: I appreciate your fix and would like more of your help with EXWM.

Copyright assignment is not needed for this change.

In any case, this ch

mgi:
medranocalvo ***@***.***> writes:

> Hello @mgi,
>
> I've read the discussion on #889.  In my opinion this fix is correct
> and necessary, although there appear to be further issues in the area.
>
> I added a new commit that reverts back to your first version modifying
> it a bit.  Please have a look and modify as you see fit.

Hi @medranocalvo,

I've just your new patch and it also works for me too.

> If you are OK with the changes (incl. commit message Thanks-To, which
> might be more personal), I propose that you rebase & commit under your
> name, if that's OK for you.

Find by me.  But I'll think I need some instructions to the "rebase &
commit under your name" part 😅

> Contributions to EXWM usually need copyright assignment to the FSF.
> Please let me know (privately if you prefer) whether you have already
> assigned it for changes to Emacs.

I already have the paperwork done for FSF.

> If not, please consider doing it, as I must require it for further
> contributions.  In other words: I appreciate your fix and would like
> more of your help with EXWM.

Thanks for the kind words.  I'll try to help AFAIC.
-- 
Manuel Giraud

leungbk:
Thanks for your work on this. I've tried the latest version of this branch, and it almost always works, save for when I am in the middle of a `consult-buffer` session and the current candidate is an EXWM buffer that is being [live-previewed](https://github.com/minad/consult#live-previews). 

I can work around this for now by disabling live preview for any `consult` command that might raise an EXWM buffer.

leungbk:
> Find by me. But I'll think I need some instructions to the "rebase & commit under your name" part 😅

In the Magit transient popup for committing (`c`), the `-A` flag allows you to commit under a different name. Alternatively, you might be able to modify the name of the author in the course of an interactive rebase by editing the name of the author in the `COMMIT_EDITMSG` buffer (haven't tested this).

Stebalien:
This goes a bit haywire with miniframe because it's not an EXWM workspace. One solution is to just iterate over all frames:

```elisp
          (dolist (f (frame-list))
            (let ((strid (number-to-string id)))
              (when (string-equal strid (frame-parameter f 'outer-window-id))
                (x-focus-frame f t)))))))
```

mgi:
Steven Allen ***@***.***> writes:

> This goes a bit haywire with miniframe because it's not an EXWM
> workspace. One solution is to just iterate over all frames:

Do you mean with the mini-frame package from melpa?
-- 
Manuel Giraud

Stebalien:
Ah, yes.

mgi:
Ok.  I've just tested with mini-frame-mode and yes, it does not work at
all.  In my setup, the completion frame ends up under the xterm (for
instance) and so is invisible.  I've replace `exwm-workspace--list` with
`(frame-list)` in the current version of #890 (as you suggested) but it
does not work for me.

I agree with Adrián that this patch is not complete (as it does not work
with every package under the sun) but it could applied as it is a
progress on current emacs anyway.
-- 
Manuel Giraud

medranocalvo:
@mgi: merged!  Thank you very much.  I'll wait for #867 before making the release.

Indeed there is much to do for Emacs 29, this is another step in that direction.

@Stebalien, I thought child frames were not supported in EXMW; do you use `mini-frame-standalone`?  In any case, please open a new ticket (just copy-paste the above) and tag it with emacs-29.

@leungbk, please open a new ticket as well.

Stebalien:
> do you use mini-frame-standalone?

Yes. I'll open a new ticket.

(but honestly, I just stopped using mini-frame entirely due to other focus issues that have nothing to do with EXWM)

medranocalvo:
@mgi: 0.27 has been released, featuring your changes.  Thank you very much!

mgi:
Thank you!
-- 
Manuel Giraud

PR #894, opened 2022-11-05
systemtray: fix theme change hook

Stebalien:
- The hook is passed an argument.
- We need to force a tray icon update.


medranocalvo:
Incorporated your fixes to #867.  Thank you.

PR #897, opened 2022-11-11
Pick the correct line height for the systemtray

Stebalien:
Otherwise, we can pick up a line height from the wrong line/window.


medranocalvo:
As a tangent, this issue is pervasive in the codebase; also with `current-buffer`, `selected-window`, `current-workspace`...  Thank you for the correction.  I'm going to adjust the commit message and merge.

medranocalvo:
Merged, thank you very much.

PR #907, opened 2023-03-07
More selectively ignore top and bottom struts

jollm:
Hi,

Thanks so much to the maintainer of this package!

The intent of this change is for exwm to accommodate dock panel windows in vertically arranged multi-display environments. So far I've tested the change with polybar and an as yet unreleased version of exlybar.

The current behavior is for struts to occlude workareas in cases where a dock panel is placed at the top of a lower display or at the bottom of an upper display. This is a known issue that window managers have to work around for X. See e.g. https://blog.martin-graesslin.com/blog/2016/08/panels-on-shared-screen-edges/

## A way to A/B test:
Vertically stack some displays with xrandr, then place a dock panel such as polybar on displays in turn, observing the behavior for different workareas each time. It is sufficient to re-eval the old/new version of `exwm-workspace--update-workareas` prior to re-launching the panel.

## commit synopsis (same as the commit message)

For the case of vertical columns of workareas, top and bottom struts should only apply to workareas containing the respective edge of the strut offset.

To simplify, imagine three monitors arranged vertically and one workarea per display:
```
  +-------+
  |   1   | 2000x1000 px
  +-------+
  |   2   | 2000x1000 px
  +-------+
  |   3   | 2000x1000 px
  +-------+
```
In sexp form: ((0 0 2000 1000) (0 1000 2000 1000) (0 2000 2000 1000)) where each element represents a workarea as x, y, width, height

And example struts of the form (offset-type offset (x-start x-end)) 
- **a.** (top 42 (0 2000))
- **b.** (top 1042 (0 2000))
- **c.** (top 2042 (0 2000))
- **d.** (bottom 42 (0 2000))
- **e.** (bottom 1042 (0 2000))
- **f.** (bottom 2042 (0 2000))

Workareas adjusted for struts before this change:
- **a.** ((0 42 2000 958) (0 1000 2000 1000) (0 2000 2000 1000))
- **b.** ((0 1042 2000 -42) (0 1042 2000 958) (0 2000 2000 1000))
- **c.** ((0 2042 2000 -1042) (0 2042 2000 -42) (0 2042 2000 958))
- **d.** ((0 0 2000 1000) (0 1000 2000 1000) (0 2000 2000 902))
- **e.** ((0 0 2000 1000) (0 1000 2000 902) (0 2000 2000 -98))
- **f.** ((0 0 2000 902) (0 1000 2000 -98) (0 2000 2000 -1098))

Note that a. and d. are sensible, while b., c., e., and f. are quite user unfriendly.

After this change, the same adjusted workareas are:
- **a.** no change
- **b.** ((0 0 2000 1000) (0 1042 2000 958) (0 2000 2000 1000))
- **c.** ((0 0 2000 1000) (0 1000 2000 1000) (0 2042 2000 958))
- **d.** no change
- **e.** ((0 0 2000 1000) (0 1000 2000 902) (0 2000 2000 1000))
- **f.** ((0 0 2000 902) (0 1000 2000 1000) (0 2000 2000 1000))

The intent is to allow dock type windows such as typical status bars to occupy space in a workarea on any of a set of vertically arranged displays without occluding the other workareas due to the limitations of the X spec regarding strut offsets.


medranocalvo:
Dear @jollm,

welcome to EXWM. Thank you for this improvement, and for the detailed and clear exposition.

The first question that came into my mind while reviewing the changes was: what does the standard say?  I looked into https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45582154996464 and found:

> Struts MUST be specified in root window coordinates, that is, they are *not* relative to the edges of any view port or Xinerama monitor.

Ouch.

I looked at what other WMs do.  This is what I got quickly (my comments might be not accurate):

i3: only supports top & bottom docks, only supports _NET_WM_STRUT
qtile: Supports _NET_WM_STRUT_PARTIAL, no special support for Xinerama/Xrandr
KWin: now we are onto something: https://invent.kde.org/plasma/kwin/-/blob/master/src/workspace.cpp#L2244(search for HACK:).
I also found this message: https://mail.gnome.org/archives/wm-spec-list/2009-November/msg00005.html, which indicates that XMonad observe(s|d) Xinerama screens.
Finally, I found https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/22, a proposal to amend EWMH to support requesting structs relative to particular areas (did not read in detail).
See also https://muktupavels.id.lv/posts/better-multi-monitor-support-in-gnome-flashback
https://mail.gnome.org/archives/wm-spec-list/2021-June/msg00000.html

[…]

I wrote the above a couple of days ago while reviewing the commit; I’ve now read the PR comment above, where you already discuss the matter… sigh.

We should indeed fix this issue, and like others, ignore the spec, which as far as one can tell won’t improve.  I wonder (pie in the sky) whether the small WMs should try to cooperate on the freedesktop X11 specs, given that the desktops apparently won’t.

If I understand correctly, with your patch we will support placing docks on horizontal edges between vertical monitors, but not on vertical edges between horizontally positioned monitors.

I confess that I experienced a “magic eye” effect while reviewing this patch.  Aren’t all those pcase branches almost equal?  It would be good to factor it out in a helper function.

[…]

I had a go at it.  It got a little out of hand... Please carefully review the changes and test the changes (particularly the last two commits): I don't use struts nor have multiple monitors.

I adjusted your commit message adding a couple of links, the ChangeLog blurb and the copyright exepmtion.  Please have a look and let me know if it's OK.

By the way, I had a look at exlybar.  I’m impressed. Please, add it to https://github.com/ch11ng/exwm/wiki#third-party-extensions.  Does it support transparency?  I ask because we recently lost support for transparency in exwm-systemtray and could use a hand implementing it.  Does exlybar support acting as a system tray?

Another question I have is whether you assigned copyright for your work on Emacs to the FSF?  (It’s not strictly necessary for merging this work, I already added the copyright exeption, but can revert it if you confirm your assignment.)

jollm:
Hi,

> If I understand correctly, with your patch we will support placing docks on horizontal edges between vertical monitors, but not on vertical edges between horizontally positioned monitors.

Yes, that's true, my patch doesn't support the horizontal edges, I will try with your changes

> I had a go at it. It got a little out of hand... Please carefully review the changes and test the changes (particularly the last two commits): I don't use struts nor have multiple monitors.
> I adjusted your commit message adding a couple of links, the ChangeLog blurb and the copyright exepmtion. Please have a look and let me know if it's OK.

Thanks for doing that!, I will test soon and let you know.

> Another question I have is whether you assigned copyright for your work on Emacs to the FSF? (It’s not strictly necessary for merging this work, I already added the copyright exeption, but can revert it if you confirm your assignment.)

I don't have a copyright assignment

> By the way, I had a look at exlybar. I’m impressed. Please, add it to https://github.com/ch11ng/exwm/wiki#third-party-extensions. Does it support transparency? I ask because we recently lost support for transparency in exwm-systemtray and could use a hand implementing it. Does exlybar support acting as a system tray?

Thanks, I will add it. Currently it doesn't provide any configuration options for transparency, but it is planned. There is a system tray module

Thanks for the thorough review!

jollm:
Hi again, everything looks fine in my testing. I tried both horizontal and vertical layouts, with the dock panel in all of the possible locations top and bottom and the workareas all remain visible.

medranocalvo:
Thank you for following up.  I'll merge and publish a new EXWM version as time permits.

> > Another question I have is whether you assigned copyright for your work on Emacs to the FSF? (It’s not strictly necessary for merging this work, I already added the copyright exeption, but can revert it if you confirm your assignment.)
> 
> I don't have a copyright assignment

Please consider assigning copyright to the FSF for Emacs-related contributions, otherwise I won't be able to accept further contributions in the future.  I'd like more contributions from you in the future! Have a look a https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html for the process or ask here or in [emacs-devel@gnu.org](mailto:emacs-devel@gnu.org).  Depending on where you live and your situation it might be quick or more... paperworky.

medranocalvo:
Merged into master.  Thank you very much!

medranocalvo:
Just released 0.28, which includes this improvement.  Thank you!

Issue #922, opened 2023-09-09
The commit 67c5b31 introduce unexpected behavior when calling "exwm-workspace-toggle-minibuffer"


zhenhua-wang:
Emacs main frame does not resize correctly after merging this commit.

When `exwm-workspace-minibuffer-position` is `bottom`, `exwm-workspace-toggle-minibuffer` would leave a line of blank space at the bottom (minibuffer position).

zhenhua-wang:
I find out that the `exwm-geometry` does not get updated after `(exwm-workspace-detach-minibuffer)`

zhenhua-wang:
As a temporary fix, I force `exwm-geometry` to have screen width and height before updating workareas

``` elisp
(defun exwm-workspace-detach-minibuffer ()
  "Detach the minibuffer so that it automatically hides."
  (interactive)
  (exwm--log)
  (when (and (exwm-workspace--minibuffer-own-frame-p)
             (exwm-workspace--minibuffer-attached-p))
    (setq exwm-workspace--attached-minibuffer-height 0)
    (let ((container (frame-parameter exwm-workspace--minibuffer
                                      'exwm-container)))
      (dolist (f exwm-workspace--list)
        (set-frame-parameter f 'exwm-geometry
                             (make-instance 'xcb:RECTANGLE
                                            :x 0
                                            :y 0
                                            :width (x-display-pixel-width)
                                            :height (x-display-pixel-height))))
      (setq exwm-workspace--id-struts-alist
            (assq-delete-all container exwm-workspace--id-struts-alist))
      (exwm-workspace--update-struts)
      (exwm-workspace--update-workareas)
      (dolist (f exwm-workspace--list)
        (exwm-workspace--set-fullscreen f))
      (exwm-workspace--hide-minibuffer))))
```

medranocalvo:
Thank you for the report and the workaround.  I think that I found the cause (an error in 67c5b316b).  Please, test latest master (8cab6c03fa2cc59d5863f5cc67197d480452c22a).

medranocalvo:
By the way, thank you for testing latest EXWM.   I'm looking for people willing to test specific areas of EXWM, please see https://github.com/ch11ng/exwm/wiki/Maintenance#exwms-testing-team.  Would you like to enroll for the *minibuffer* and *general* (or other) areas?  (To be clear, I don't think we'd have catched this one earlier...)

zhenhua-wang:
Thanks for the quick fix. The latest master works very well!

[...]

PR #923, opened 2023-09-09
Fix frame height after calling exwm-workspace-detach-minibuffer

zhenhua-wang:
This solves the issue #922 by restoring the minibuffer height within the main Emacs frame.


zhenhua-wang:
Fixed in [8cab6c0](https://github.com/ch11ng/exwm/commit/8cab6c03fa2cc59d5863f5cc67197d480452c22a)

PR #935, opened 2023-12-28
Implement the XSETTINGS protocol

### LINKED ISSUE #876: Change GTK Themes in EXWMj
I would like to create my own function to toggle between light and dark themes. However, I would also like to toggle between GTK themes (adwaita -> adwaita-dark). Hopefully firefox will recognize the GTK theme and also adjust accordingly.

However, i can't find a way to change the GTK Theme outside of gnome. I tried editing the .config/gtk* files, the gsettings and the gtk-chtheme and gtk-theme-swtich packages. Nothing changed my themes for Evince or Nautilus.

I am using Debian experimental with emacs 28.1

Thanks

Stebalien:
Users can use this to configure system-wide themes, icons, fonts, etc.

```elisp
(setopt exwm-xsettings-theme '("Adwaita" . "Adwaita-dark")
        exwm-xsettings-icon-theme '("Papirus-Light" . "Papirus-Dark")
        exwm-xsettings '(("Gtk/FontName" . "IBM Plex Sans 9")
                         ("Gtk/DocumentFontName" . "IBM Plex Sans 9")
                         ("Gtk/MonospaceFontName" . "JetBrains Mono 9")))
```

fixes #876

NOTE: If enabled, this module will conflict with any other XSETTINGS implementation and will refuse to start.

---

* exwm-xsettings.el: Implement the XSETTINGS protocol (fixes #876)

### DIFF
--- a/exwm-xsettings.el
+++ b/exwm-xsettings.el
@@ -21,8 +21,8 @@
 
 ;;; Commentary:
 
-;; Implements the XSETTINGS protocol, allowing Emacs to manage the
-;; system theme, fonts, icons, etc.
+;; Implements the XSETTINGS protocol, allowing Emacs to manage the system theme,
+;; fonts, icons, etc.
 ;;
 ;; This package can be configured as follows:
 ;;
@@ -39,9 +39,10 @@
 ;;                          ("Xft/Hinting" . 1)))
 ;;   (exwm-xsettings-enable)
 ;;
-;; To modify these settings at runtime, customize them with `custom-set-variables' or `setopt'
-;; (Emacs 29+). E.g., the following will immediately change the icon theme to "Papirus" at runtime,
-;; even in running applications:
+;; To modify these settings at runtime, customize them with
+;; `custom-set-variables' or `setopt' (Emacs 29+). E.g., the following will
+;; immediately change the icon theme to "Papirus" at runtime, even in running
+;; applications:
 ;;
 ;;   (setopt exwm-xsettings-icon-theme "Papirus")
 
@@ -70,7 +71,8 @@ SYMBOL is the setting being updated and VALUE is the new value."
 
 (defcustom exwm-xsettings nil
   "Custom XSETTINGS.
-These settings take precedence over `exwm-xsettings-theme' and `exwm-xsettings-icon-theme'."
+These settings take precedence over `exwm-xsettings-theme' and
+`exwm-xsettings-icon-theme'."
   :group 'exwm
   :type '(alist :key-type (string :tag "Name")
                 :value-type (choice :tag "Value"
--- a/exwm-xsettings.el
+++ b/exwm-xsettings.el
@@ -1,6 +1,6 @@
 ;;; exwm-xsettings.el --- XSETTINGS Module for EXWM -*- lexical-binding: t -*-
 
-;; Copyright (C) 2022 Free Software Foundation, Inc.
+;; Copyright (C) 2022-2024 Free Software Foundation, Inc.
 
 ;; Author: Steven Allen <steven@stebalien.com>
 
--- a/exwm-xsettings.el
+++ b/exwm-xsettings.el
@@ -47,6 +47,9 @@
 
 ;;; Code:
 
+(require 'xcb-ewmh)
+(require 'xcb-icccm)
+
 (require 'exwm-core)
 
 (defvar exwm-xsettings--connection nil)
@@ -166,6 +169,15 @@ These settings take precedence over `exwm-xsettings-theme' and `exwm-xsettings-i
    (green :initarg :green :type xcb:CARD16)
    (blue :initarg :blue :type xcb:CARD16)
    (alpha :initarg :alpha :initform #xffff :type xcb:CARD16)))
+
+(defclass xcb:xsettings:-ClientMessage
+  (xcb:icccm:--ClientMessage xcb:ClientMessage)
+  ((format :initform 32)
+   (type :initform 'xcb:Atom:MANAGER)
+   (time :initarg :time :type xcb:TIMESTAMP)      ;new slot
+   (selection :initarg :selection :type xcb:ATOM) ;new slot
+   (owner :initarg :owner :type xcb:WINDOW))      ;new slot
+  :documentation "An XSETTINGS client message.")
 
 (defun exwm-xsettings--pick-theme (theme)
   "Pick a light or dark theme from the given THEME.
@@ -329,8 +341,7 @@ SERIAL is a sequence number."
                        :destination exwm--root
                        :event-mask xcb:EventMask:StructureNotify
                        :event (xcb:marshal
-                               ;; TODO: new event? Do I need a new field?
-                               (make-instance 'xcb:systemtray:-ClientMessage
+                               (make-instance 'xcb:xsettings:-ClientMessage
                                               :window exwm--root
                                               :time xcb:Time:CurrentTime
                                               :selection exwm-xsettings--XSETTINGS_S0-atom
--- a/exwm-xsettings.el
+++ b/exwm-xsettings.el
@@ -1,0 +1,343 @@
+;;; exwm-xsettings.el --- XSETTINGS Module for EXWM -*- lexical-binding: t -*-
+
+;; Copyright (C) 2022 Free Software Foundation, Inc.
+
+;; Author: Steven Allen <steven@stebalien.com>
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Implements the XSETTINGS protocol, allowing Emacs to manage the
+;; system theme, fonts, icons, etc.
+
+;;; Code:
+
+(require 'exwm-core)
+
+(defvar exwm-xsettings--connection nil)
+(defvar exwm-xsettings--XSETTINGS_SETTINGS-atom nil)
+(defvar exwm-xsettings--XSETTINGS_S0-atom nil)
+(defvar exwm-xsettings--selection-owner-window nil)
+
+(defun exwm-xsettings--rgba-match (_widget value)
+  "Return t if VALUE is a valid RGBA color."
+  (and (numberp value) (<= 0 value 1)))
+
+(defcustom exwm-xsettings nil
+  "Custom XSETTINGS.
+These settings take precedence over `exwm-xsettings-theme' and `exwm-xsettings-icon-theme'."
+  :group 'exwm
+  :type '(alist :key-type (string :tag "Name")
+                :value-type (choice :tag "Value"
+                              (string :tag "String")
+                              (integer :tag "Integer")
+                              (list :tag "Color"
+                                (number :tag "Red"
+                                        :type-error
+                                        "This field should contain a number between 0 and 1."
+                                       :match exwm-xsettings--rgba-match)
+                                (number :tag "Green"
+                                        :type-error
+                                        "This field should contain a number between 0 and 1."
+                                       :match exwm-xsettings--rgba-match)
+                                (number :tag "Blue"
+                                        :type-error
+                                        "This field should contain a number between 0 and 1."
+                                       :match exwm-xsettings--rgba-match)
+                                (number :tag "Alpha"
+                                        :type-error
+                                        "This field should contain a number between 0 and 1."
+                                       :match exwm-xsettings--rgba-match
+                                       :value 1.0))))
+  :initialize #'custom-initialize-default
+  :set (lambda (symbol value)
+         (set-default-toplevel-value symbol value)
+         (exwm-xsettings--update-settings)))
+
+(defcustom exwm-xsettings-theme nil
+  "The system-wide theme."
+  :group 'exwm
+  :type '(choice (string :tag "Theme")
+                 (cons (string :tag "Light Theme")
+                       (string :tag "Dark Theme"))))
+
+(defcustom exwm-xsettings-icon-theme nil
+  "The system-wide icon theme."
+  :group 'exwm
+  :type '(choice (string :tag "Icon Theme")
+                 (cons (string :tag "Light Icon Theme")
+                       (string :tag "Dark Icon Theme"))))
+
+(defvar exwm-xsettings--serial 0)
+
+(defconst xcb:xsettings:-Type:Integer 0)
+(defconst xcb:xsettings:-Type:String 1)
+(defconst xcb:xsettings:-Type:Color 2)
+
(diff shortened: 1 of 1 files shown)


exwm-xsettings.el
```diff
@@ -0,0 +1,343 @@
...
+                       :owner id
+                       :selection exwm-xsettings--XSETTINGS_S0-atom
+                       :time xcb:Time:CurrentTime))
+    (xcb:+request exwm-xsettings--connection
+        (make-instance 'xcb:SendEvent
+                       :propagate 0
+                       :destination exwm--root
+                       :event-mask xcb:EventMask:StructureNotify
+                       :event (xcb:marshal
+                               ;; TODO: new event? Do I need a new field?
```
Stebalien: This is one of the things I'm not sure about. I'm "borrowing" the systemtray client message struct because it's identical, but maybe I should define a new one?

exwm-xsettings.el
```diff
@@ -0,0 +1,343 @@
...
+                 (cons (string :tag "Light Icon Theme")
+                       (string :tag "Dark Icon Theme"))))
+
+(defvar exwm-xsettings--serial 0)
+
+(defconst xcb:xsettings:-Type:Integer 0)
+(defconst xcb:xsettings:-Type:String 1)
+(defconst xcb:xsettings:-Type:Color 2)
+
+(defclass xcb:xsettings:-Settings
```
Stebalien, line 126: Should I move some of this to xelb?

exwm-xsettings.el
```diff
@@ -0,0 +1,343 @@
...
+
+    ;; Detect loss of XSETTINGS ownership.
+    (xcb:+event exwm-xsettings--connection 'xcb:SelectionClear
+                #'exwm-xsettings--on-SelectionClear)
+
+    (xcb:flush exwm-xsettings--connection))
+
+  ;; Update the xsettings if/when the theme changes.
+  (add-hook 'enable-theme-functions #'exwm-xsettings--on-theme-change)
+  (add-hook 'disable-theme-functions #'exwm-xsettings--on-theme-change))
```
minad, line 375: I suggest to make the update settings function public, such that it could be called from the user configuration where appropriate instead of registering these hooks. If desired the user can still register update-settings as theme hook.

EDIT: I saw that you also include code to switch between the light and dark variants of the Gtk theme. This is nice and also what I use in my configuration right now by reloading xsettingsd. Nevertheless it may be better to decouple the theme gluing code from the xsettings functionality here, and leave the coupling to the user configuration? I still hope to find configurable Gtk (and ideally also Qt) themes somewhere, which would allow me to transfer some of the basic colors from the Emacs theme (background, foreground), instead of only toggling between Adwaita Light and Dark. If such a thing would exist, the simple dark/light toggle wouldn't be enough.

Stebalien: A user should be able to just call `(setopt exwm-xsettings-theme "NewTheme")`, they don't need to call these functions.

You're right that this code isn't, strictly speaking, necessary. Nor is it sufficient for all use-cases. But it probably covers _most_ cases and it's pretty convenient.

minad: > A user should be able to just call (setopt exwm-xsettings-theme "NewTheme"), they don't need to call these functions.

Right. Exposing the update function is not necessary.

> You're right that this code isn't, strictly speaking, necessary. Nor is it sufficient for all use-cases. But it probably covers most cases and it's pretty convenient.

Yes, having this functionality here with the possibility to opt-out should not hurt. Maybe if the `exwm-xsettings-theme` variables are not set, the functionality could be completely disabled, no hooks installed?

exwm-xsettings.el
```diff
@@ -0,0 +1,343 @@
...
+                                       :match exwm-xsettings--rgba-match)
+                                (number :tag "Alpha"
+                                        :type-error
+                                        "This field should contain a number between 0 and 1."
+                                       :match exwm-xsettings--rgba-match
+                                       :value 1.0))))
+  :initialize #'custom-initialize-default
+  :set (lambda (symbol value)
+         (set-default-toplevel-value symbol value)
+         (exwm-xsettings--update-settings)))
```
minad: Extract this setter as a separate function and also use if for `exwm-xsettings-theme` and icon theme?

Stebalien: Ah, yep. I missed that `(setop exwm-xsettings-theme ...)` doesn't actually apply immediately. Fixed.

exwm-xsettings.el
Stebalien: >  the functionality could be completely disabled, no hooks installed?

I considered it but instead made them no-ops when not required.

- The logic required for automatically adding-removing the hooks is more complex and the cost of running the hooks when they're "not used" is minimal.
- Adding them in `exwm-xsettings--init` gives the user a way to reliably remove them if they _really_ don't want this behavior.
- EXWM already hooks into `enable-theme-functions` and `disable-theme-functions` in a few places to detect changes in the background color (system tray and the background module).

minad: Thanks! Makes sense. Looking forward to your PR getting merged.

medranocalvo:
Thank you @Stebalien.  I'll review as soon as possible.

minad:
As a data point, I've installed this `exwm-xsettings` package in my configuration about a week ago. It works flawlessly (theme switching and other settings). I think the only thing I am missing here is a little bit of documentation. Maybe an example configuration in the commentary would do?

This is what I use:

~~~ elisp
(require 'exwm-xsettings)
(setq exwm-xsettings-theme '("Adwaita" . "Adwaita-dark")
      exwm-xsettings '(("Xft/HintStyle" . "hintslight")
                       ("Xft/RGBA" . "rgb")
                       ("Xft/lcdfilter" . "lcddefault")
                       ("Xft/Antialias" . 1)
                       ("Xft/Hinting" . 1)
                       ("Xft/DPI" . 147456)))
(exwm-xsettings-enable)
~~~

Stebalien:
LGTY?

exwm-xsettings.el
```diff
@@ -0,0 +1,373 @@
+;;; exwm-xsettings.el --- XSETTINGS Module for EXWM -*- lexical-binding: t -*-
+
+;; Copyright (C) 2022 Free Software Foundation, Inc.
```
medranocalvo: Is the year correct?  (No problem if it is.)

exwm-xsettings.el
medranocalvo: Hmmm, good question. (I'm thinking out loud here.)  Most (all?) definitions like this in XELB are generated from the XML protocol definitions (xcb-xproto).  The XSettings spec is not part of xcb-xproto and there does not seem to exist a machine-readable spec understandable by XELB's el_client.el.  Systemtray, which is an XDG spec just like XSettings, defines the messages in the exwm code.  I'd say it belongs here.

exwm-xsettings.el
medranocalvo: Yes, please.  Copypaste it here with a new name.  They are different messages although at this time they both look same.

medranocalvo: It seems that both are selection announcement messages (https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#manager_selections).  I think it is better to have two different messages, one for each manager.

medranocalvo:
I have reviewed the code.  I answered a couple of questions.  I'm happy to merge when you think it's ready (e.g. remove TODO: comment).

This is also a great feature, thank you very much!

exwm-xsettings.el
Stebalien: I did write it in 2022, but updated it this year so I'll make it a range.

exwm-xsettings.el
Stebalien: Fixed.

Stebalien:
Ok, I've addressed everything and it should be good to go now. Thanks for the review!

medranocalvo:
One final nit: up to now I've followed strictly the Emacs commit message conventions (https://github.com/ch11ng/exwm/wiki/Maintenance#commit-messages), mainly because @ch11ng did.  I use Magit and `C-x 4 a` and it is easy enough for me.  But maybe it's time to revise this "policy".  Should we follow Emacs commit message conventions?  @minad, I'd like you to weigh in as well.

minad:
In my projects I mostly don't follow the convention, so we could as well relax it. But if mostly people feel like keeping the convention I won't mind.

Stebalien:
So, there are two issues:

1. Commit messages.
2. Commits themselves.

With respect to the first, I kind of like Emacs's commit format as it tells you exactly what changed and why, but it's also kind of annoying. I don't have a strong preference either way (but I've updated the PR description with what the commit message _should_ be).

With respect to the second, the issue here is that GitHub doesn't have proper "patch sets". That leaves three options for addressing PR review.

1. Open a new PR (terrible).
4. Amend/rebase and push (difficult to review).
5. Keep adding new commits.

So I've gotten into the habit of doing the latter. Unfortunately, that means that the best way to merge PRs (without random commits) is, IMO, squash-merge.

---

For now, I'd:

1. Squash merge.
2. Keep the commit message format, but consider formatting the commit message (when squash-merging) on behalf of contributor if it's not too complex.

That is, unless the patch requires no changes and/or is a nicely formatted patch set. In that case, just rebase-merge.

medranocalvo:
@Stebalien I agree on both 1. and 2.

On 1., I find the consistent messages slightly helpful, it is kind of a semi-automatic summary.  The format allows the more usual commentary with the summary at the bottom (see git log for examples).  Using `C-x 4 a` from the Magit diff (it might work in VC?) makes the chore easy.  I've been editing most (?) contribution's commit messages and it's not been a big hassle for me.  I believe that we shouldn't require contributors to write it themselves.

On 2. I believe that most contributions have been single commits (not sure?).  I prefer the adding of further commits with the changes after review, but in the end it will be whatever the contributor does.  (We can follow your proposal between us.) Squashing and writing the final message seems a comfortable default.

exwm-xsettings.el
```diff
@@ -0,0 +1,384 @@
...
+(defun exwm-xsettings--custom-set (symbol value)
+  "Setter used by `exwm-xsettings' customization options.
+
+SYMBOL is the setting being updated and VALUE is the new value."
+  (set-default-toplevel-value symbol value)
+  (exwm-xsettings--update-settings))
+
+(defcustom exwm-xsettings nil
+  "Custom XSETTINGS.
+These settings take precedence over `exwm-xsettings-theme' and `exwm-xsettings-icon-theme'."
```
minad: @Stebalien We should break the long doc string lines here, since the byte compiler will otherwise complain. I have checkdoc and package-lint always enabled in my Elisp buffers which will warn me early on about such nits.

Stebalien: Load-path issues were preventing it from linting the rest of the file.

exwm-xsettings.el
```diff
@@ -0,0 +1,386 @@
...
+
+(defun exwm-xsettings--pick-theme (theme)
+  "Pick a light or dark theme from the given THEME.
+If THEME is a string, it's returned directly.
+If THEME is a cons of (LIGHT . DARK), the appropriate theme is picked based on
+the default face's background color."
+  (pcase theme
+    ((cl-type string) theme)
+    (`(,(cl-type string) . ,(cl-type string))
+     (if (color-dark-p (color-name-to-rgb (face-background 'default)))
```
Stebalien: Ok, so, this requires emacs 28.1. Should I just not support this feature on older Emacs versions?

minad: I wouldn't like if certain EXWM features are only available on newer Emacs versions. EXWM as a whole should work coherently on the supported Emacs version.

We could either implement the missing function for older Emacs versions or pull in Compat as dependency. Compat may be useful if we want to use other functionality. I cannot tell right now, but maybe this will be the case later (eg when we refactor some parts of the code).

Independent of this issue I suggest that we raise the dependency to 27.1 if there are no objections and support 27.1 for a few years. 27.1 is widely available on Linux distributions, including the ones with long time support.

minad: See https://github.com/emacs-compat/compat/blob/dff3639b70f3c9916c5397056a51cdc3e60d5668/compat-28.el#L672-L699 for the relevant code in Compat.

Stebalien: So, options are:

1. Copy some Emacs code here.
2. Depend on compat (may let us use newer emacs features elsewhere as well).
3. Remove the automatic dark mode switching (as pointed out by @minad earlier, users can handle dark-mode switching themselves).

At this point, I'm leaning towards option 3, actually. I _like_ the feature, but this is the second time we've tripped over it (and there's no reason we can't add it back in the future).

Preferences @minad, @medranocalvo?

minad: My preference is 1, 2, 3 in this order. To elaborate:

1. This is the quickest solution in the short term, and easy to replace with Compat if we start using it.
2. In the long term I propose that we depend on Compat, since this will help us using newer APIs. This can be done as part of some refactoring. Compat will also be officially supported for Emacs core packages soon (see https://lists.gnu.org/archive/html/bug-gnu-emacs/2024-01/msg00660.html).
3. Removing the feature was my preference when I first reviewed this patch. But by now I am bit more pragmatic. I use the theme switching and it really seems the easiest way to enable this out of the box for the users. It would be unfortunate if users would have to reimplement this themselves.

minad: @Stebalien How shall we proceed? I think it is okay to merge this into the current master branch to make it available to a wider user base. The addition is low risk since it must be enabled explicitly.

Stebalien: Done. Sorry, I try to limit my Emacs hacking to the weekends (try being the operative word).

exwm-xsettings.el
```diff
@@ -181,6 +181,21 @@ These settings take precedence over `exwm-xsettings-theme' and
...
+            ;; Borrowed from Emacs 29.
+            (lambda (rgb)
+              "Whether RGB is more readable against white than black."
+              (unless (<= 0 (apply #'min rgb) (apply #'max rgb) 1)
+                (error "RGB components %S not in [0,1]" rgb))
+              (let* ((r (expt (nth 0 rgb) 2.2))
+                     (g (expt (nth 1 rgb) 2.2))
+                     (b (expt (nth 2 rgb) 2.2))
+                     (y (+ (* r 0.2126) (* g 0.7152) (* b 0.0722))))
+                (< y 0.325)))
```
Stebalien: I shrunk this code down a bit and inlined the constant (we're not maintaining this code here, so readability isn't a goal.

exwm-xsettings.el
```diff
@@ -181,6 +181,21 @@ These settings take precedence over `exwm-xsettings-theme' and
    (owner :initarg :owner :type xcb:WINDOW))      ;new slot
   :documentation "An XSETTINGS client message.")
 
+(eval-and-compile
+  (fset 'exwm-xsettings--color-dark-p
+        (if (< emacs-major-version 29)
```
minad: I suggest to write this in a slightly different form with `defalias`, which I consider cleaner:

~~~ elisp
(defalias 'exwm-xsettings--color-dark-p 
   (if (eval-when-compile (< emacs-major-version 29)) ...))
~~~

exwm-xsettings.el
minad: Inlining the constant makes sense.

exwm-xsettings.el
Stebalien: Fixed and amended.

exwm-xsettings.el
minad: Okay, thanks! Since we cannot merge here, feel free to push this to https://github.com/emacs-exwm/exwm and close the PR here. I just installed the current version of exwm-xsettings.el in my configuration and gave it a final test.

Stebalien:
Moved to https://github.com/emacs-exwm/exwm/pull/2. I kept it on my local fork for now, although I'll probably start pushing branches directly to emacs-exwm/exwm in the future.

Issue #570, opened 2019-04-20
Transfer focus to emacs for org-resolve-clocks-if-idle


shader:
If I'm using an X application, like Firefox, and the org idle timer resolution window pops up, the resolution input characters are eaten by Firefox and do not tell org how to resolve the clock.

However, if I click on the other window to transfer focus and resolve the clock, emacs receives not only the focus event, but also a mouse click. This causes an error, because the org-resolve-clocks-if-idle is expecting a character input event.

Any ideas on how to transfer focus without clicking?

Using "C-x o" to switch windows doesn't work either (even though normally C-x would not be captured by firefox), presumably because the org resolution is checking only for its specific input characters, and ignoring all others.

ch11ng:
The problem is `org-clock-resolve` gets user input with `read-char` and it's difficult to detect it (there is no active minibuffer).  `org-clock-resolve` does specify a timeout of 45s but typically users won't wait for that long.

> Using "C-x o" to switch windows doesn't work either (even though normally C-x would not be captured by firefox), presumably because the org resolution is checking only for its specific input characters, and ignoring all others.

Yes, it doesn't work even if you do `org-resolve-clocks` without any `exwm-mode` buffer.

medranocalvo:
@shader, this should be fixed now.  Could you test the master branch?

PR #938, opened 2023-12-29
Passthrough input when explicitly reading events and keys

### LINKED ISSUE #921: `org-resolve-clocks` (and other spontaneous focus-grabing functions) has conflict with EXWM-mode buffers
While working in an EXWM-mode buffer if one clocks in an Org mode task, org consider user being idle and call org-resolve-clocks. This results in EXWM sending all normal keys (like k or i which are needed to deal with org-resolve-clocks) to the EXWM-mode buffer so one can not interact with `org-resolve-clocks` and all EXWM related keys (win+1..7 for example) is being taken by `org-resolve-clocks` so one can not interact with EXWM (to get out of EXWM-mode buffer). 

The only way out seems to be either quit the app in EXWM buffer (if the app offer a way to do that) or send `abort-recursive-edit` from another tty.

Stebalien:
When any of these functions are called, Emacs expects to receive all input events until they return. Unfortunately, there doesn't appear to be a reliable way to detect that these functions are currently running other than advising them, so that's what I did.

These functions are:

- `read-char`
- `read-char-exclusive`
- `read-key-sequence-vector`
- `read-key-sequence`
- `read-event`

In terms of downsides... I've been advising these functions (well, most of them) in my config for years now and haven't had any issues.

Fixes #921, #570 (and probably others)

### DIFF
--- a/exwm-input.el
+++ b/exwm-input.el
@@ -102,6 +102,13 @@ defined in `exwm-mode-map' here."
 (defconst exwm-input--update-focus-interval 0.01
   "Time interval (in seconds) for accumulating input focus update requests.")
 
+(defconst exwm-input--passthrough-functions '(read-char
+                                              read-char-exclusive
+                                              read-key-sequence-vector
+                                              read-key-sequence
+                                              read-event)
+  "Low-level functions that read events and need to be exempted from EXWM's input handling.")
+
 (defvar exwm-input--during-command nil
   "Indicate whether between `pre-command-hook' and `post-command-hook'.")
 
@@ -1158,6 +1165,11 @@ One use is to access the keymap bound to KEYS (as prefix keys) in `char-mode'."
     (exwm--log)
     (exwm-input--on-minibuffer-exit)))
 
+(defun exwm-input--call-with-passthrough (function &rest args)
+  "Bind `exwm-input-line-mode-passthrough' and call the specified FUNCTION with ARGS."
+  (let ((exwm-input-line-mode-passthrough t))
+    (apply function args)))
+
 (defun exwm-input--init ()
   "Initialize the keyboard module."
   (exwm--log)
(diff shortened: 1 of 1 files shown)


exwm-input.el
```diff
@@ -102,6 +102,13 @@ defined in `exwm-mode-map' here."
 (defconst exwm-input--update-focus-interval 0.01
   "Time interval (in seconds) for accumulating input focus update requests.")
 
+(defconst exwm-input--passthrough-functions '(read-char
```
Stebalien, line 105: I've defined this as a private constant instead of as a `defcustom` as this list _should_ be exhaustive and I figured we could make it public later.

medranocalvo: I tend to agree.  We can revisit if someone would like it to be public.

medranocalvo:
This is great, thank you.  We already advise when there's no other way.  (Ideally, we should report Emacs bugs for stuff we advise, in case a mechanism exists or can be made for us...)

I'll merge and I'd propose leaving this some time in master for people to try it out.  Does that sound OK?

minad:
medranocalvo ***@***.***> writes:

> This is great, thank you. We already advise when there's no other way. (Ideally,
> we should report Emacs bugs for stuff we advise, in case a mechanism exists or
> can be made for us...)

Yes, we should do that - at least for the cases we consider particularly
intrusive/problematic. Personally I am not fond of the indirection via
the server for blocking GUI features.

> I'll merge and I'd propose leaving this some time in master for people to try it
> out. Does that sound OK?

Sounds good. I am installing EXWM from GNU ELPA-devel. If this is not
yet the case we should document this installation method, such that more
users can beta test new functionality.

medranocalvo:
> medranocalvo ***@***.***> writes:
> This is great, thank you. We already advise when there's no other way. (Ideally, we should report Emacs bugs for stuff we advise, in case a mechanism exists or can be made for us...)
> Yes, we should do that - at least for the cases we consider particularly intrusive/problematic. Personally I am not fond of the indirection via the server for blocking GUI features.

I opened to track that https://github.com/ch11ng/exwm/issues/943 (we should do the migration and avoid opening more tickets over here.  I'll look into working on that on the following days, though you two are welcome push forward).

> I'll merge and I'd propose leaving this some time in master for people to try it out. Does that sound OK?
> Sounds good. I am installing EXWM from GNU ELPA-devel. If this is not yet the case we should document this installation method, such that more users can beta test new functionality.

There's many users that track the master branch here on GitHub.  I think elpa-devel is less well known and we should encourage it.  In https://github.com/ch11ng/exwm/wiki/Maintenance there's also two ideas I planned but never actually brought into action: EXWM Testing Team, people who opt in to test particular features (think systemtray or XRandr, which I don't use); and Call For Testing, kind of a release candidate announcement where the testing team members would be @tagged.

medranocalvo:
I merged this to master and pushed to externals/exwm as well.  Congrats!  (Soon you'll merge yourself :-)).

PR #939, opened 2023-12-30
Correctly stop the subordinate Emacs daemon

Stebalien:
This patch starts Emacs as a foreground daemon (so it can accurately be killed) and terminates it by sending a signal (because `server-force-delete` stops the local server, not the named server).

### DIFF
--- a/exwm.el
+++ b/exwm.el
@@ -5,7 +5,7 @@
 ;; Author: Chris Feng <chris.w.feng@gmail.com>
 ;; Maintainer: Adrián Medraño Calvo <adrian@medranocalvo.com>
 ;; Version: 0.28
-;; Package-Requires: ((xelb "0.18"))
+;; Package-Requires: ((emacs "26.1") (xelb "0.18"))
 ;; Keywords: unix
 ;; URL: https://github.com/ch11ng/exwm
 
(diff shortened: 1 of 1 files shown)


exwm.el
```diff
@@ -1017,7 +1025,7 @@ FUNCTION is the function to be evaluated, ARGS are the arguments."
                          (car command-line-args) ;The executable file
                          "-d" (frame-parameter nil 'display)
                          "-Q"
-                         (concat "--daemon=" exwm--server-name)
+                         (concat "--fg-daemon=" exwm--server-name)
```
Stebalien, line 1028: This requires Emacs 26, but I think that's safe at this point.

minad: Maybe require Emacs 26 version here: https://github.com/ch11ng/exwm/blob/56db521a3ed6a2ccc1193110a56545a1992e9654/exwm.el#L8

medranocalvo:
> because server-force-delete stops the local server, not the named server

This sounded like a bug, but might be an API deficiency.  Would let-binding `server-name` allow `server-start` to work?

```elisp
(let ((server-name "our EXWM subordinate server name"))
  (server-stop)
```

Stebalien:
Unfortunately, no. Emacs stops the server running in the current process regardless of the name and never actually tries to communicate with any remote servers.

medranocalvo:
I was confused: I thought we started a second server on EXWM's Emacs. No: we start a new Emacs and start the server there.  It makes no sense trying to use `server-stop`.

I'll merge soon.

medranocalvo:
This has been merged.  Thank you.

PR #937, opened 2023-12-29
exwm-blocking-subrs: Add message-box and message-or-box (Fix #874)

minad:

PR #945, opened 2024-01-09
exwm-blocking-subrs: x-* functions may be missing on some Emacs builds

minad:
Avoid compiler warnings. Follow-up of #937.

### DIFF
--- a/exwm.el
+++ b/exwm.el
@@ -98,8 +98,9 @@
   :type 'hook)
 
 (defcustom exwm-blocking-subrs
-  (list #'x-file-dialog #'x-popup-dialog #'x-select-font
-        #'message-box #'message-or-box)
+  ;; `x-file-dialog' and `x-select-font' are missing on some Emacs builds, for
+  ;; example on the X11 Lucid build.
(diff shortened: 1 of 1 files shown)


exwm.el
```diff
@@ -98,7 +98,9 @@
   :type 'hook)
 
 (defcustom exwm-blocking-subrs
-  (list #'x-file-dialog #'x-popup-dialog #'x-select-font
+  ;; `x-file-dialog' and `x-select-font' are missing on some Emacs builds, for
+  ;; example Lucid.
+  (list 'x-file-dialog 'x-popup-dialog 'x-select-font
         #'message-box #'message-or-box)
```
minad: @medranocalvo @Stebalien If preferred, we can also go back to a quoted list here instead of using hashed symbols.

Stebalien: I'd remove the hash from all of them, for consistency.

minad: Done.

minad:
Merged into emacs-exwm/exwm: https://github.com/emacs-exwm/exwm/commit/c069e8195dfdd81f458e4a0dac4d9589077321e3

PR #946, opened 2024-01-09
Version related fixes

minad:
I suggested in https://github.com/ch11ng/exwm/pull/935#discussion_r1446362095 to bump the Emacs dependency to 27.1. These days, 27.1 is widely available (for users and testers, 28 not yet). It seems like a reasonable minimal dependency. We can also remove a tiny bit of code, specific to Emacs 26. How do you feel about this?

### DIFF
--- a/exwm-floating.el
+++ b/exwm-floating.el
@@ -31,7 +31,6 @@
 
 (defgroup exwm-floating nil
   "Floating."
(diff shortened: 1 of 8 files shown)


minad:
Moved to https://github.com/emacs-exwm/exwm/pull/1 for discussion.

PR #947, opened 2024-02-05
Moved to https://github.com/emacs-exwm/exwm

medranocalvo:
@minad, please have a look.  I mostly copied your wording from the wiki, but removed our names because if I would disappear there would be no way to update them (in contrast to the wiki, which is editable by anyone).


minad (approved):
LGTM!

